PR #68, opened 2024-04-20
TsFile v4 to support table-interfaces

jt2594838:
In order to support manipulating TsFile in a table-like (or relational-like) manner for upper-level systems like IoTDB, new interfaces like `registerTable` and `writeTable` are introduced into TsFile.

Also, deviceIds are fully generic by replacing the instances with the interface IDeviceID everywhere. A similar refactor is done to MeasurementSchema.

Experiments indicate the throughput downgrade on write is small (< 3%) while the query time consumption using the table-view can be reduced by 80%. The increment in file size is negligible (< 0.1%).


tsfile/src/main/java/org/apache/tsfile/file/metadata/StringArrayDeviceID.java
```diff
@@ -0,0 +1,271 @@
...
+  public long ramBytesUsed() {
+    return INSTANCE_SIZE + RamUsageEstimator.sizeOf(segments);
+  }
+
+  @Override
+  public int serializedSize() {
+    int cnt = Integer.BYTES;
+    for (String segment : segments) {
+      cnt += Integer.BYTES;
+      cnt += segment.getBytes(TSFileConfig.STRING_CHARSET).length;
```
JackieTien97: segment may be null here.

tsfile/src/main/java/org/apache/tsfile/file/metadata/StringArrayDeviceID.java
```diff
@@ -0,0 +1,271 @@
...
+    return 0;
+  }
+
+  @Override
+  public long ramBytesUsed() {
+    return INSTANCE_SIZE + RamUsageEstimator.sizeOf(segments);
+  }
+
+  @Override
+  public int serializedSize() {
```
JackieTien97: while serializing string, we use varInt to represent string length and string array count, may only occupy just one byte.

tsfile/src/main/java/org/apache/tsfile/file/metadata/StringArrayDeviceID.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+    return new StringArrayDeviceID(segments);
+  }
+
+  @Override
+  public byte[] getBytes() {
+    ByteArrayOutputStream publicBAOS = new ByteArrayOutputStream(256);
+    for (String segment : segments) {
+      try {
+        publicBAOS.write(segment.getBytes(TSFileConfig.STRING_CHARSET));
```
JackieTien97: segement may be null here.

JackieTien97 (changes requested):

PR #76, opened 2024-05-13
Support new data type: STRING, BLOB, TIMESTAMP, DATE

Cpaulyz:
Support new data type: STRING, BLOB, TIMESTAMP, DATE

https://apache-iotdb.feishu.cn/docx/VHnMdyW6WobOrZx3o46cuFF9nJd

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/utils/DateUtils.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/utils/DateUtils.java
@@ -28,11 +28,11 @@ public class DateUtils {
   private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd");
 
   public static String formatDate(int date) {
-    String dateStr = String.valueOf(date);
-    String year = dateStr.substring(0, 4);
-    String month = dateStr.substring(4, 6);
-    String day = dateStr.substring(6, 8);
-    return year + "-" + month + "-" + day;
+    return date / 10000
+        + "-"
+        + String.format("%02d", (date / 100) % 100)
+        + "-"
+        + String.format("%02d", date % 100);
   }
 
   public static Integer parseDateExpressionToInt(String dateExpression) {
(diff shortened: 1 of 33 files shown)


java/tsfile/src/main/java/org/apache/tsfile/utils/DateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+    String day = dateStr.substring(6, 8);
+    return new Date(
+        Integer.parseInt(year) - 1900, Integer.parseInt(month) - 1, Integer.parseInt(day));
+  }
+
+  public static LocalDate parseIntToLocalDate(int date) {
+    String dateStr = String.valueOf(date);
+    String year = dateStr.substring(0, 4);
+    String month = dateStr.substring(4, 6);
+    String day = dateStr.substring(6, 8);
```
JackieTien97: use `/` and `%` to calculate will be better?

java/tsfile/src/main/java/org/apache/tsfile/utils/DateUtils.java
```diff
@@ -0,0 +1,84 @@
...
+    return localDate.getYear() * 10000
+        + localDate.getMonthValue() * 100
+        + localDate.getDayOfMonth();
+  }
+
+  public static Date parseIntToDate(int date) {
+    String dateStr = String.valueOf(date);
+    String year = dateStr.substring(0, 4);
+    String month = dateStr.substring(4, 6);
+    String day = dateStr.substring(6, 8);
```
JackieTien97: use / and % to calculate will be better?

JackieTien97 (changes requested):

java/tsfile/src/main/java/org/apache/tsfile/utils/DateUtils.java
Cpaulyz: Done. And I add a UT for it.

java/tsfile/src/main/java/org/apache/tsfile/utils/DateUtils.java
Cpaulyz: Done. And I add a UT for it.

JackieTien97 (approved):

PR #77, opened 2024-05-14
Fix TsBlockBuilder bug in AlignedPageReader and PageReader.

JackieTien97:
We should only use limit to init TsBlockBuilder while we truly have limit value.

In some use cases, we can see query improvement with this pr.
from 5s -> 700ms

Because we save the array copy and memory estimate in the processing of array growing.


liuminghui233 (approved):
LGTM :-)

PR #78, opened 2024-05-16
feat: Support interface in device level of RestorableTsFileIOWriter

ycycse:
In the past, we only accessed TsFile in terms of series as the basic unit, and the existing implementation has adapted well to this. However, in the future, we may introduce larger granularity scanning units, which would require a TsFile access interface at the device level.

This PR introduces an interface to RestorableTsFileIOWriter to get chunkMetaData in device level.

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/write/writer/RestorableTsFileIOWriter.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/write/writer/RestorableTsFileIOWriter.java
@@ -197,18 +197,29 @@ public class RestorableTsFileIOWriter extends TsFileIOWriter {
     return chunkMetadataList;
   }
 
-  public List<ChunkMetadata> getVisibleMetadataList(IDeviceID deviceId, TSDataType dataType) {
-    List<ChunkMetadata> chunkMetadataList = new ArrayList<>();
+  /**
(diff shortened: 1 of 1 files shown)


java/tsfile/src/main/java/org/apache/tsfile/write/writer/RestorableTsFileIOWriter.java
```diff
@@ -186,12 +185,26 @@ public List<ChunkMetadata> getVisibleMetadataList(
...
         if (dataType == null || dataType.equals(chunkMetaData.getDataType())) {
-          chunkMetadataList.add((ChunkMetadata) chunkMetaData);
+          chunkMetadataList.add(chunkMetaData);
+        }
+      }
+    }
+    return chunkMetadataList;
+  }
+
+  public List<ChunkMetadata> getVisibleMetadataList(IDeviceID deviceId, TSDataType dataType) {
```
JackieTien97:
```suggestion
  public List<List<ChunkMetadata>> getVisibleMetadataList(IDeviceID deviceId, TSDataType dataType) {
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #79, opened 2024-05-17
Add copy slice methods to all Column classes.

Sh-Zh-7:
Add `getRegionCopy()` and `subColumnCopy()` methods which return slice copy instead of its view.

### DIFF
--- a/java/tsfile/src/test/java/org/apache/tsfile/read/common/ColumnTest.java
+++ b/java/tsfile/src/test/java/org/apache/tsfile/read/common/ColumnTest.java
@@ -45,20 +45,42 @@ public class ColumnTest {
 
   @Test
   public void timeColumnSubColumnTest() {
-    TimeColumnBuilder columnBuilder = new TimeColumnBuilder(null, 10);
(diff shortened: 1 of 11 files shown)


java/tsfile/src/test/java/org/apache/tsfile/read/common/ColumnTest.java
```diff
@@ -45,20 +45,42 @@ public class ColumnTest {
 
   @Test
   public void timeColumnSubColumnTest() {
-    TimeColumnBuilder columnBuilder = new TimeColumnBuilder(null, 10);
+    LongColumnBuilder columnBuilder = new LongColumnBuilder(null, 10);
```
JackieTien97: change back to `TimeColumnBuilder`

JackieTien97 (changes requested):

java/tsfile/src/test/java/org/apache/tsfile/read/common/ColumnTest.java
Sh-Zh-7: Done.

JackieTien97 (approved):

PR #80, opened 2024-05-17
Return all columns in TsBlock class

Sh-Zh-7:
Expand value columns in place and put time column at the last of the array.

This new method is needed in UDTF executor framework.

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/read/common/block/TsBlock.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/read/common/block/TsBlock.java
@@ -215,6 +215,7 @@ public class TsBlock {
 
   /**
    * Collected all columns into a column array. Note that the timestamps are at the last column.
+   *
    * @return column array composed of all columns.
    */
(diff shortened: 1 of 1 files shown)


java/tsfile/src/main/java/org/apache/tsfile/read/common/block/TsBlock.java
```diff
@@ -213,6 +213,12 @@ public Column[] getColumns(int[] columnIndexes) {
     return columns;
   }
 
+  public Column[] getAllColumns() {
```
JackieTien97, line 221: add java doc for this method, specially for its return value

JackieTien97 (changes requested):

java/tsfile/src/main/java/org/apache/tsfile/read/common/block/TsBlock.java
Sh-Zh-7: Done.

JackieTien97 (approved):

PR #97, opened 2024-06-04
Tsfile cwrapper

ColinLeeo:
In this pr:
Add cwrapper to tsfile cpp. User can use tsfile in C env.
Add some examples to tsfile_cpp examples.
refine CMakelists


JackieTien97 (changes requested):
Add some guidelines in README about C wrapper.

PR #110, opened 2024-06-13
feature: finish multiple flush for c++_tsfile

RkGrit:
feature: finish multiple flush for c++_tsfile


cpp/src/common/global.cc
```diff
@@ -50,6 +50,8 @@ void init_config_value() {
     g_config_value_.page_writer_max_memory_bytes_ = 128 * 1024;  // 128 k
     g_config_value_.max_degree_of_index_node_ = 256;
     g_config_value_.tsfile_index_bloom_filter_error_percent_ = 0.05;
+    g_config_value_.record_count_for_next_mem_check_ = 100;
+    g_config_value_.chunk_group_size_threshold_ = 128 * 1024;
```
jt2594838: It is not reasonable enough to set the size of a chunk group to 128 KB when the max memory of a single page is also 128 KB. At least it should be one or two magnitudes higher.

cpp/src/encoding/bitpack_encoder.h
```diff
@@ -58,9 +58,7 @@ class BitPackEncoder {
         packer_ = nullptr;
     }
 
-    void destroy() { /* do nothing for BitPackEncoder */
-        delete (packer_);
-    }
+    void destroy() { /* do nothing for BitPackEncoder */ delete (packer_); }
```
jt2594838: Why does it say "do nothing"?

cpp/src/encoding/bitpack_encoder.h
```diff
@@ -167,6 +165,24 @@ class BitPackEncoder {
         bytes_buffer_.clear();
         bitpacked_group_count_ = 0;
     }
+
+    int get_max_byte_size() {
+        if (values_.empty()) {
+            return 0;
+        }
+        int maxBitWidth = get_int_max_bit_width(values_);
```
jt2594838: Optional: the max_bit_width can be calculated in a streaming manner so that you will not have to calculate it every time.

cpp/src/encoding/bitpack_encoder.h
```diff
@@ -167,6 +165,24 @@ class BitPackEncoder {
...
+        int totalValues = values_.size();
+        int fullGroups = totalValues / 8;
+        int remainingValues = totalValues % 8;
+        int bytesPerGroup = (maxBitWidth * 8 + 7) / 8;
+        int maxSize = 0;
+        maxSize += fullGroups * bytesPerGroup;
+        if (remainingValues > 0) {
+            maxSize += bytesPerGroup;
+        }
+        maxSize += fullGroups * (1 + 1) + (remainingValues > 0 ? (1 + 1) : 0);
```
jt2594838, line 174: Please add some comments for this (1 + 1).

cpp/src/encoding/dictionary_encoder.h
```diff
@@ -95,6 +95,11 @@ class DictionaryEncoder {
     void write_encoded_data(common::ByteStream &out) {
         values_encoder_.encode_flush(out);
     }
+
+    int get_max_byte_size() 
+    {
+        return 4 + map_size_ + values_encoder_.get_max_byte_size();
```
jt2594838, line 102: Please also add comments here.

I notice that the DictionaryEncoder uses a BitPackingEncoder as the inner encoder. However, in the Jave edition, the inner encoder is an RleEncoder (with bit-packing). This may cause incompatibility.

cpp/src/encoding/encoder.h
```diff
@@ -38,6 +38,14 @@ class Encoder {
...
     virtual int encode(double value, common::ByteStream &out_stream) = 0;
     virtual int flush(common::ByteStream &out_stream) = 0;
+
+    /**
+     * The maximal possible memory size occupied by current Encoder. This
+     * statistic value doesn't involve OutputStream.
+     *
+     * @return the maximal size of possible memory occupied by current encoder
+     */
+    virtual int get_max_byte_size() = 0;
```
jt2594838, line 48: The implementation in BitPackEncoder is not the memory size because the memory data is not encoded; it calculates the encoded size.

cpp/src/encoding/gorilla_encoder.h
```diff
@@ -120,6 +120,7 @@ class GorillaEncoder : public Encoder {
     }
 
     int get_one_item_max_size();
+    int get_max_byte_size();
```
jt2594838: What is the difference between the two? Their implementations are identical.

cpp/src/file/tsfile_io_writer.cc
```diff
@@ -169,7 +169,7 @@ int TsFileIOWriter::flush_chunk(ByteStream &chunk_data) {
         // log_err("flush stream error, ret=%d", ret);
     }
 
-    chunk_data.destroy();
+    // chunk_data.destroy();
     return ret;
```
jt2594838: Remove it.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -220,8 +220,7 @@ int TsFileWriter::do_check_schema(const std::string &device_name,
         if (UNLIKELY(ms_iter == msm.end())) {
             chunk_writers.push_back(NULL);
         } else {
-            // Here we may check data_type against ms_iter. But in Java
-            // libtsfile, no check here.
+            // In Java we will check data_type. But in C++, no check here.
```
jt2594838, line 228: Please give a reason for this.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -242,6 +241,44 @@ int TsFileWriter::do_check_schema(const std::string &device_name,
...
+    for (device_iter = schemas_.begin(); device_iter != schemas_.end();
+         device_iter++) {
+        MeasurementSchemaGroup *chunk_group = device_iter->second;
+        MeasurementSchemaMap &map = chunk_group->measurement_schema_map_;
+        for (MeasurementSchemaMapIter ms_iter = map.begin();
+             ms_iter != map.end(); ms_iter++) {
+            MeasurementSchema *m_schema = ms_iter->second;
+            ChunkWriter *&chunk_writer = m_schema->chunk_writer_;
+            mem_total_size += chunk_writer->estimate_max_series_mem_size();
+        }
```
jt2594838: Is there a chance that chunk_writer is null?

cpp/src/writer/tsfile_writer.cc
```diff
@@ -242,6 +241,44 @@ int TsFileWriter::do_check_schema(const std::string &device_name,
...
+}
+
+/**
+ * check occupied memory size, if it exceeds the chunkGroupSize threshold, flush
+ * them to given OutputStream.
+ * 
+ * @return true - size of tsfile or metadata reaches the threshold. false -
+ * otherwise
+ */
+int TsFileWriter::check_memory_size_and_may_flush_chunks() {
```
jt2594838: I do not think the return value is boolean.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -393,15 +433,35 @@ int TsFileWriter::flush() {
...
+         device_iter++) {
+
+        if (check_chunk_group_empty(device_iter->second)) {
+            continue;
+        }
+
         if (RET_FAIL(io_writer_->start_flush_chunk_group(device_iter->first))) {
         } else if (RET_FAIL(flush_chunk_group(device_iter->second))) {
         } else if (RET_FAIL(io_writer_->end_flush_chunk_group())) {
         }
```
jt2594838: If the return value indicates an error, it should be returned immediately.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -419,6 +479,8 @@ int TsFileWriter::flush_chunk_group(MeasurementSchemaGroup *chunk_group) {
                        chunk_writer->get_chunk_data()))) {
         } else if (RET_FAIL(io_writer_->end_flush_chunk(
                        chunk_writer->get_chunk_statistic()))) {
+        } else {
+            chunk_writer->destroy();
         }
```
jt2594838: If the return value indicates an error, it should be returned immediately.

jt2594838 (changes requested):
Good job, but there are some minor issues.

ColinLeeo:
Some UT tests should be added currently.

cpp/src/writer/page_writer.h
```diff
@@ -134,6 +134,18 @@ class PageWriter {
...
         return time_out_stream_.total_size() + value_out_stream_.total_size();
     }
+    /**
+     * calculate max possible memory size it occupies, including time
+     * outputStream and value outputStream, because size outputStream is never
+     * used until flushing.
+     *
+     * @return allocated size in time, value and outputStream
+     */
+    FORCE_INLINE uint32_t estimate_max_mem_size() const {
```
ColinLeeo, line 144: TsFile 超过4GB 之后返回值可能会泄露？

cpp/src/writer/tsfile_writer.cc
ColinLeeo: 这里现在应该还是lazy 初始化的，如果注册了序列，未写入，chunk_writer 就是null

jt2594838: So, line 255 may use a null pointer.
The nullability of chunk_writer should be checked.

cpp/src/common/global.cc
RkGrit: Ok, I'll change "chunk_group_size_threshold_" to 128 * 1024 * 1024

cpp/src/encoding/bitpack_encoder.h
RkGrit: This is a comment left over by history, which was initially intended to indicate that the destroy method requires no special cleanup, and just the removal of packer_.

RkGrit: I will delete the comment

cpp/src/encoding/bitpack_encoder.h
RkGrit: Ok, I will optimize this place.

cpp/src/encoding/bitpack_encoder.h
RkGrit: "1+1" means each bitpack group has a header of 1 byte and a tail of 1 byte

cpp/src/encoding/dictionary_encoder.h
RkGrit: In c++ edition, we don't hava RleEncoder and the DictionaryEncoder is not in use currently.

cpp/src/encoding/gorilla_encoder.h
RkGrit: It's redundant. I'll delete get_one_item_max_size();

cpp/src/file/tsfile_io_writer.cc
RkGrit: OK

cpp/src/writer/tsfile_writer.cc
RkGrit: because checks are performed at the chunk layer and page layer

cpp/src/writer/tsfile_writer.cc
RkGrit: OK

cpp/src/writer/tsfile_writer.cc
RkGrit: I will fix the comments

cpp/src/writer/tsfile_writer.cc
RkGrit: Ok

cpp/src/writer/tsfile_writer.cc
RkGrit: OK

cpp/src/encoding/bitpack_encoder.h
jt2594838: Add in the code also, not just the conversation.

cpp/src/writer/tsfile_writer.cc
jt2594838: Add in the code also, not just the conversation.

cpp/src/writer/tsfile_writer.h
```diff
@@ -119,7 +119,7 @@ class TsFileWriter {
     int64_t record_count_since_last_flush_;
     // record count for next memory check
     int64_t record_count_for_next_mem_check_;
-    bool start_file_done_;
+    // bool start_file_done_;
```
jt2594838: Remove it.

jt2594838 (approved):

cpp/src/writer/tsfile_writer.h
RkGrit: ok

cpp/src/writer/tsfile_writer.cc
RkGrit: ok

PR #112, opened 2024-06-14
Python wrapper.

ColinLeeo:


python/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+                            <goal>wget</goal>
+                        </goals>
+                        <configuration>
+                            <url>https://github.com/ThrowTheSwitch/Unity/archive/v${unity.version}.zip</url>
+                            <unpack>true</unpack>
+                            <outputDirectory>${project.build.directory}/dependency</outputDirectory>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
```
chrisdutz: Do we really need this? Unity is a C test framework.

chrisdutz:
Possibly also worth having a look how the folks at PLC4X did the Python integration into the build for PLC4Py https://github.com/apache/plc4x/blob/develop/plc4py/pom.xml (They are using venv, which I think is a good idea in the Python world)

python/pom.xml
ColinLeeo: Ok I will remove this.

ColinLeeo:
> Possibly also worth having a look how the folks at PLC4X did the Python integration into the build for PLC4Py https://github.com/apache/plc4x/blob/develop/plc4py/pom.xml (They are using venv, which I think is a good idea in the Python world)

OK，I will try it.

python/setup.py
```diff
@@ -0,0 +1,98 @@
...
+        language="c++"
+    )
+]
+
+setup(
+    name="tsfile",
+    version="0.1",
+    description="tsfile reader and writer for python",
+    url="https://github.com/apache/tsfile",
+    author="Colin Lee",
```
HTHou: I checked the code of the IoTDB python client. The author is Apache Software Foundation.

python/setup.py
```diff
@@ -0,0 +1,98 @@
...
+        library_dirs=[libtsfile_dir],
+        include_dirs=[include_dir, np.get_include()],
+        runtime_library_dirs=[libtsfile_dir],
+        extra_compile_args=["-std=c++11"],
+        language="c++"
+    )
+]
+
+setup(
+    name="tsfile",
```
HTHou, line 104: I think using "apache-tsfile" as name is better.

python/tsfile/tsfile.py
```diff
@@ -0,0 +1,139 @@
...
+    column_names = data.columns.tolist()
+    column_types = data.dtypes
+
+    if TIMESTAMP_STR not in column_names:
+        raise AttributeError("Time column is missing")
+    if column_types[TIMESTAMP_STR] != "int64":
+        raise TypeError("Time column must be of type int64")
+    allowed_types = {"int64", "int32", "bool", "float32", "float64"}
+
+    # 检查每列的数据类型是否在允许的类型集合中
```
HTHou: Using english here?

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,359 @@
...
+
+    
+
+
+
+
+
+        
+        
+
```
HTHou: Remove these empty lines?

python/tsfile/tsfile_pywrapper.cpp
HTHou, line 1: Is this file generated by Cython? If so, I think we can generate it by maven in the `generate-sources` phase. Better not include it directly in source code.

python/setup.py
ColinLeeo: 这里的包对应的是发布的包名称，tsfile 比较直接

HTHou: I remember the ASF requires the name of python package starting with `apache-`. 
See https://incubator.apache.org/guides/distribution.html#pypi 

It that true? @chrisdutz

.github/workflows/unit-test.yml
HTHou, line 1:
``` 
     - name: Install python packages
        if: ${{ matrix.os == 'ubuntu-latest'}}
        run: pip3 install cython pandas numpy
```

This part can be replace like that?
```
      - name: Install python packages
        run: pip3 install requirement.txt
```

ColinLeeo: Oh，this part can be deleted. Installation can be done by maven.

python/setup.py
ColinLeeo: https://github.com/apache/plc4x/blob/develop/plc4py/setup.py#L23
Well, in plcpy, the name is plc4py

python/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+        <!-- Tell Sonar where to find the sources -->
+        <sonar.sources>tsfile</sonar.sources>
+        <sonar.cfamily.build-wrapper-output>${project.build.directory}/build-wrapper-output</sonar.cfamily.build-wrapper-output>
+    </properties>
+    <build>
+        <sourceDirectory>${project.basedir}</sourceDirectory>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
```
HTHou, line 42: Another issue, you may need to add this dependency in the main pom to fix a build warning.
Like this https://github.com/apache/tsfile/pull/122

ColinLeeo: Ok, I will check it

python/setup.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+project_dir = os.path.dirname(__file__)
+libtsfile_shard_dir = os.path.join(project_dir, "..", "cpp", "target", "build", "lib")
+libtsfile_dir = os.path.join(project_dir, "tsfile")
+include_dir = os.path.join(project_dir, "tsfile")
+source_file = os.path.join(project_dir, "tsfile", "tsfile_pywrapper.pyx")
+
+if platform.system() == "Darwin":
+    copy_lib_files("Darwin", libtsfile_shard_dir, libtsfile_dir, "dylib", "1.0")
```
HTHou:
```suggestion
    copy_lib_files("Darwin", libtsfile_shard_dir, libtsfile_dir, "dylib")
```

python/README.md
```diff
@@ -0,0 +1,72 @@
...
+Using pylint to check Python code is recommended. However, there is no suitable style checking tool for Cython code, and this part of the code should be consistent with the Python style required by pylint.
+
+**Feature List**
+- [ ] In pywrapper, invoke the batch reading interface implemented in tsfile_CPP.
+
+
+
+## Build
+
+You can build tsfile_python by mvn or shell command.
```
jt2594838: Better to use consistent terms and styles. 
You have used "The Python version of TsFile," "TsFile CPP," and "tsfile_python."
In this case, it may seem less confusing. 
However, in longer articles, indicating the same conception with different terms can cause reading difficulty.

jt2594838: "by mvn or shell command."
I only find "mvn" and "python" commands below, so where is "shell."
Well, technically, they are both shell commands.
Maybe "by mvn or python command."

python/README.md
```diff
@@ -0,0 +1,72 @@
...
+Build by mvn in this directory:
+
+```sh
+mvn initialize compile test
+```
+
+Build by mvn in tsfile root directory:
+
+```sh
+mvn clean package -P with-python
```
jt2594838: Why not provide the command as you write in unit-test.yml, which also includes the compilation of TsFile-CPP.
I mean, it is more convenient since the user does not have to switch between files.

python/examlpes.py
```diff
@@ -0,0 +1,79 @@
...
+import numpy as np
+import pandas as pd
+import os
+
+import tsfile as ts
+
+
+# test writing data
+data_dir = os.path.join(os.path.dirname(__file__), "test.tsfile")
+TABLE_NAME = "test_table"
```
jt2594838: Is the table model now supported by TsFile-CPP? If not, better not to use such concepts.

python/examlpes.py
```diff
@@ -0,0 +1,79 @@
...
+
+# 1000 rows data
+time = np.arange(1, 1001, dtype=np.int64)
+level = np.linspace(2000, 3000, num=1000, dtype=np.float32)
+num = np.arange(10000, 11000, dtype=np.int64)
+df = pd.DataFrame({"Time": time, "level": level, "num": num})
+
+if os.path.exists(data_dir):
+    os.remove(data_dir)
+ts.write_tsfile(data_dir, TABLE_NAME, df)
```
jt2594838: So, one TsFile can only be written with one DataFrame.
This may be a great limitation.
Maybe we should have a further discussion on interfaces.

python/examlpes.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# read data we already wrote
+# with 20 chunksize
+tsfile_ret = ts.read_tsfile(data_dir, TABLE_NAME, ["level", "num"], chunksize=20)
+print(tsfile_ret.shape)
+
+# # with 100 chunksize
+tsfile_ret = ts.read_tsfile(data_dir, TABLE_NAME, ["level", "num"], chunksize = 100)
+print(tsfile_ret.shape)
```
jt2594838: The code styles seem a little in consistent for "chunksize".
Can the spotless plugin be applied to python?

python/examlpes.py
```diff
@@ -0,0 +1,79 @@
...
+
+# # get all data
+tsfile_ret = ts.read_tsfile(data_dir, TABLE_NAME, ["level", "num"])
```
jt2594838: Is it true that each call will open a new reader so the metadata cannot be reused?
This may have some issues with efficiency, especially with intensive analysis.
We can delay the optimization, but better to find somewhere to record this issue (leave a todo).

python/examlpes.py
```diff
@@ -0,0 +1,79 @@
...
+tsfile_ret = ts.read_tsfile(data_dir, TABLE_NAME,
+                             ["level"], start_time=50, end_time=100, chunksize=10)
+print(tsfile_ret.shape)
+
+# with time scale
+tsfile_ret = ts.read_tsfile(data_dir, TABLE_NAME, ["num"], start_time=50, end_time=100)
+print(tsfile_ret.shape)
+
+
+with ts.read_tsfile(
```
jt2594838: Missing a comment here, since each example above is with one.

python/setup.py
```diff
@@ -0,0 +1,125 @@
...
+
+project_dir = os.path.dirname(__file__)
+libtsfile_shard_dir = os.path.join(project_dir, "..", "cpp", "target", "build", "lib")
+libtsfile_dir = os.path.join(project_dir, "tsfile")
+include_dir = os.path.join(project_dir, "tsfile")
+source_file = os.path.join(project_dir, "tsfile", "tsfile_pywrapper.pyx")
+
+if platform.system() == "Darwin":
+    copy_lib_files("Darwin", libtsfile_shard_dir, libtsfile_dir, "dylib")
+    copy_lib_files("Darwin", libtsfile_shard_dir, libtsfile_dir, "1.0", "dylib")
```
jt2594838: Should the last two parameters be reverted? I think the last one is the version.

python/test.py
```diff
@@ -0,0 +1,155 @@
...
+    df = pd.DataFrame({"Time": time, "level": level, "num": num})
+    with ut.TestCase().assertRaises(TypeError):
+        ts.write_tsfile(DATA_PATH + "/wrong_time_type.tsfile", TABLE_NAME, df)\
+        
+    # TXT is not support yet
+    time = np.arange(1, 1001, dtype=np.int64)
+    text = np.random.choice(["a", "b", "c"], 1000)
+    df = pd.DataFrame({"Time": time, "text": text})
+    with ut.TestCase().assertRaises(TypeError):
+        ts.write_tsfile(DATA_PATH + "/txt.tsfile", TABLE_NAME, df)
```
jt2594838, line 66: This can be important for some applications.
Do we have a todo list for such missing functionalities?

python/test.py
```diff
@@ -0,0 +1,155 @@
...
+    ts.write_tsfile(DATA_PATH + "/full_datatypes.tsfile", TABLE_NAME, df)
+
+# test reading data
+def test_read_tsfile():
+    # skip test on windows because of the bug in the tsfile library
+    if platform.system() == "Windows":
+        return
+    # test read not exist file
+    with ut.TestCase().assertRaises(FileNotFoundError):
+        ts.read_tsfile(DATA_PATH + "/notexit.tsfile", TABLE_NAME, ["level", "num"])
```
jt2594838: notexit->notexist

python/test.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # test read empty file
+    with open(DATA_PATH + "/empty.tsfile", "w", encoding="utf-8") as f:
+        pass
+
+    with ut.TestCase().assertRaises(ValueError):
+        ts.read_tsfile(DATA_PATH + "/empty.tsfile", TABLE_NAME, ["level", "num"])
```
jt2594838: Can this ValueError be replaced with something more specific?

python/tsfile/tsfile.py
```diff
@@ -0,0 +1,142 @@
...
+    allowed_types = {"int64", "int32", "bool", "float32", "float64"}
+
+    for col, dtype in column_types.items():
+        if dtype.name not in allowed_types:
+            raise TypeError(
+                f"Column '{col}' has an invalid type '{dtype}'."
+            )
+
+    writer = tsfile_writer(file_path)
+    writer.write_tsfile(table_name, data)
```
jt2594838, line 145: Will the writer (and the reader above) be properly closed?

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,330 @@
...
+        c_table_name = PyBytes_AsString(py_table_name)
+        if isinstance(columns, str):
+            columns = [columns]
+        
+        c_columns = <char**>malloc(len(columns) * sizeof(char*))
+        if not c_columns:
+            raise MemoryError("Failed to allocate memory for columns")
+
+        for i in range(len(columns)):
+            c_columns[i] = <char*>malloc(len(columns[i]) + 1)
```
jt2594838, line 84: Where will  c_columns be freed?

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,330 @@
...
+                raise MemoryError("Failed to allocate memory for columns")
+            column_binary = columns[i].encode('utf-8')
+            column = PyBytes_AsString(column_binary)
+            strcpy(c_columns[i], column)
+        # query data from tsfile
+        if start_time is not None or end_time is not None:
+            if start_time is None:
+                start_time = -1
+            if end_time is None:
+                end_time = -1
```
jt2594838: Is this correct? Timestamps may be negative.

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,330 @@
...
+        else:
+            self.ret = ts_reader_read(self.reader, table_name.encode('utf-8'), c_columns, len(columns))
+        
+        
+    def read_tsfile(self):
+        # open tsfile to read
+        res = pd.DataFrame()
+        if self.batch_size == -1:
+            self.batch_size = 1024
+            while True:
```
jt2594838: Is this mentioned anywhere in the interface? The default batch_size is 1024.

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,330 @@
...
+                arr_bool_ = np.empty(length, dtype=np.bool_)
+                bool_ptr = <char*> result.value[i]
+                for j in range(length):
+                    arr_bool_[j] = bool_ptr[j] != 0
+                arr = np.array(arr_bool_, dtype = np.bool_)
+
+            elif schema.column_def  == TS_TYPE_FLOAT:
+                np_array_float = cnp.PyArray_SimpleNewFromData(1, &length, cnp.NPY_FLOAT32, result.value[i])
+                arr = np.array(np_array_float, dtype = np.float32)
+                arr = np.where(is_not_null, arr, np.nan)
```
jt2594838, line 218: Are you trying to use NaN to mark null values? This may be problematic because the original value can be NaN.

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,330 @@
...
+            
+            if has_null:
+                tmp_array = np.full(length, np.nan, np.float64)
+                tmp_array[is_not_null] = arr[is_not_null]
+                if schema.column_def == TS_TYPE_INT32:
+                    arr = pd.Series(tmp_array).astype('Int32')
+                elif schema.column_def == TS_TYPE_BOOLEAN:
+                    arr = pd.Series(tmp_array).astype(np.bool_)
+                elif schema.column_def == TS_TYPE_INT64:
+                    arr = pd.Series(tmp_array).astype('Int64')
```
jt2594838, line 243: What is this part for?

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,330 @@
...
+        for i in range(len(df)):
+            time = df.iloc[i][TIMESTAMP_STR]
+            self.create_row_data(table_name, time, len(column_names))
+            for j in range(1, len(column_names)):
+                column_name = column_names[j]
+                column_value = df.iloc[i][column_name]
+                column_ctype = column_ctypes[j]
+                self.write_into_row_data(column_name, column_value, column_ctype)
+            if tsfile_write_row_data(self.writer, self.row_data) != 0:
+                raise Exception("Failed to write row data")
```
jt2594838, line 336: Writing a tablet (DataFrame) in a row-based manner also has efficiency issues. We should add a todo for this.

jt2594838 (changes requested):

python/examlpes.py
ColinLeeo: 这里的table就是timeseries的逻辑。在设计C接口的时候也讨论过timeseries 包装为table逻辑。便于理解。

python/examlpes.py
ColinLeeo: Further feature.需要支持多次flush

python/examlpes.py
ColinLeeo: 元数据复用带来的收益不高，查询时tsfile 中的元数据也是按需加载的。

python/test.py
ColinLeeo: 这里应该归入TsFile cpp的todo中。

python/tsfile/tsfile.py
ColinLeeo: writer 是python 中的对象，python 处理，在writer 写完数据之后，会将自己拿到的C++的资源释放掉。

python/tsfile/tsfile_pywrapper.pyx
ColinLeeo: 内部标记，指示该值不存在。

python/tsfile/tsfile_pywrapper.pyx
ColinLeeo: 这不是默认的batch_size, 这是一个内部的结构，按照这样大的size 生成中间的dataframe，这些dataframe 会组装起来。

python/tsfile/tsfile_pywrapper.pyx
ColinLeeo: 原始数据不会是NAN，这是C接口返回的结构体。

python/tsfile/tsfile_pywrapper.pyx
ColinLeeo: 需要支持返回的数据中的NAN，int  和bool 类型不支持设置NaN，需要转换之后才可以。同 AINode 中的ainode/iotdb/ainode/parser.py 处理逻辑。

.github/workflows/unit-test.yml
```diff
@@ -73,13 +73,16 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12' 
 
       # Run the actual maven build including all unit- and integration-tests.
       - name: Build and test with Maven (All others)
         shell: bash
         run: |
-          ./mvnw${{ steps.platform_suffix.outputs.platform_suffix }} -P with-java,with-cpp clean verify
+          ./mvnw${{ steps.platform_suffix.outputs.platform_suffix }} -P with-cpp,with-python clean verify
```
HTHou: Now we can add `with-java` back.

.gitignore
```diff
@@ -16,3 +16,11 @@ docs/node_modules/
...
 docs/src/.vuepress/.temp/
 docs/src/.vuepress/dist/
+
+# python files
+python/build
+python/tsfile/__pycache__
+python/tsfile/*.so*
+python/tsfile/TsFile_cwrapper.h
+python/tsfile/*.cpp
+python/data
```
HTHou: Should add `dll`, `dylib`, `venv`?

python/setup.py
HTHou: I tried to build it on my mac, the name of the lib file is `libtsfile.1.0.dylib`. I think `copy_lib_files` method should consider it.

python/test.py
```diff
@@ -0,0 +1,155 @@
...
+
+import unittest as ut
+import numpy as np
+import pandas as pd
+
+if platform.system() == "Windows":
+    extra_dll_dir = os.path.join(os.path.dirname(__file__), "tsfile")
+    os.add_dll_directory(extra_dll_dir)
+    print(extra_dll_dir)
+    print(glob.glob(extra_dll_dir + '/*'))
```
HTHou: These printing can be removed.

python/examlpes.py
jt2594838: Aside from the previous tree model in TsFile V3, we will present a table model in TsFile V4, where the table-related concepts will be truly added.
Before then, please use consistent concepts like Device and Measurement.

python/setup.py
jt2594838: Then, the argument names should be reconsidered to avoid confusion.

python/tsfile/tsfile_pywrapper.pyx
jt2594838: What if the user really passes -1 as the parameter?

python/tsfile/tsfile_pywrapper.pyx
jt2594838: Why can the original data not be NaN? I can write NaNs using the writing interface and the TsFile should store them.

python/tsfile/tsfile_pywrapper.pyx
jt2594838: If it is not a default value, then why do you set it to 1024 when batch_size == -1 ?

python/tsfile/tsfile_pywrapper.pyx
jt2594838: int and bool do not have NaN and do not need to return NaN.
I am afraid that you are mixing NaN with Null, which are totally different.

python/tsfile/tsfile.py
jt2594838: There are `def __dealloc__(self):` and `def __exit__(self, exc_type, exc_value, traceback):` for tsfile_reader, but only `def __exit__(self, exc_type, exc_value, traceback):` for tsfile_writer.
Is there any specific reason that tsfile_writer does not need a `def __dealloc__(self):`?

python/examlpes.py
jt2594838: Not making sense. The TsFileMetadata and high-level index nodes can be highly reused. Let alone there are multiple caches in the Java edition.

python/examlpes.py
jt2594838: Well, it is not even “multiple flushes“. It is “multiple insertions . “

python/tsfile/tsfile.py
ColinLeeo: __dealloc__ 方法是应对with open 语法的，writer没有这个方法。

python/tsfile/tsfile_pywrapper.pyx
ColinLeeo: 这是个内部的结构，作用于返回的tablet，表示C接口一次查询返回的内存大小。1024 可以随意设置，这只是方便减少拷贝次数做的。

python/tsfile/tsfile_pywrapper.pyx
ColinLeeo: Well， if the result is NULL, you cannot return a dataframe with `NULL`, you have to return a dataframe  with NaN.

.gitignore
ColinLeeo: Yes, you are right.

python/examlpes.py
ColinLeeo: Added Todo list.

python/examlpes.py
ColinLeeo: Well, Spotless does not support Python. 
I use black formatter to format python code.

jt2594838:
image
Please avoid using force-push.

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -0,0 +1,341 @@
...
+                tmp_array[is_not_null] = arr[is_not_null]
+                if schema.column_def == TS_TYPE_INT32:
+                    arr = pd.Series(tmp_array).astype('Int32')
+                elif schema.column_def == TS_TYPE_BOOLEAN:
+                    arr = pd.Series(tmp_array).astype(np.bool_)
+                elif schema.column_def == TS_TYPE_INT64:
+                    arr = pd.Series(tmp_array).astype('Int64')
+
+            res[column_name] = arr
+            null_map.append(is_not_null)
```
jt2594838: null_map or not_null_map?

jt2594838 (approved):

python/tsfile/tsfile_pywrapper.pyx
```diff
@@ -86,41 +90,52 @@ cdef class tsfile_reader:
...
     def __iter__(self):
         return self
     
     def __next__(self):
-        res = self.get_next_dataframe()
+        res, null_map = self.get_next_dataframe()
         if res is None:
             raise StopIteration
-        return res
+        return res, null_map
```
jt2594838: not_null_map?

HTHou (approved):

jt2594838 (approved):

PR #128, opened 2024-06-19
Added non aligned query to tsFile Reader

Caideyipi:


java/tsfile/src/main/java/org/apache/tsfile/read/query/dataset/NonAlignedDataSetWithoutTimeGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+  private final List<AbstractFileSeriesReader> readers;
+  private BatchData batchData;
+
+  public NonAlignedDataSetWithoutTimeGenerator(
+      final List<Path> paths,
+      final List<TSDataType> dataTypes,
+      final List<AbstractFileSeriesReader> readers) {
+    super(paths, dataTypes);
+    this.readers = readers;
+    this.columnNum = 1;
```
jt2594838, line 43: Why is columnNum 1?

java/tsfile/src/main/java/org/apache/tsfile/read/query/dataset/NonAlignedDataSetWithoutTimeGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @Override
+  public boolean hasNextWithoutConstraint() {
+    return batchData.hasCurrent();
```
jt2594838, line 48: What if the caller calls `hasNextWithoutConstraint()` before calling `nextWithoutConstraint()`? The `batchData` is null then.

jt2594838 (changes requested):
Please also add unit tests.

PR #179, opened 2024-07-20
[CPP] Support alternative gtest url

hongzhi-gao:
1. Support alternative GTEST URL for users in China
2. Unit tests will be skipped if the download fails

### DIFF
--- a/cpp/test/CMakeLists.txt
+++ b/cpp/test/CMakeLists.txt
@@ -17,12 +17,45 @@ specific language governing permissions and limitations
 under the License.
 ]]
 include(FetchContent)
-FetchContent_Declare(
-  googletest
-  URL https://github.com/google/googletest/archive/refs/tags/v1.12.0.zip
+
+set(URL_LIST
+    "https://github.com/google/googletest/archive/refs/tags/v1.12.0.zip"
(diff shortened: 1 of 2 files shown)


cpp/test/CMakeLists.txt
```diff
@@ -17,12 +17,45 @@ specific language governing permissions and limitations
...
 ]]
 include(FetchContent)
-FetchContent_Declare(
-  googletest
-  URL https://github.com/google/googletest/archive/refs/tags/v1.12.0.zip
+
+set(URL_LIST
+    "https://github.com/google/googletest/archive/refs/tags/v1.12.0.zip"
+    "https://hub.nuaa.cf/google/googletest/archive/refs/tags/v1.12.0.zip"
+    "https://hub.yzuu.cf/google/googletest/archive/refs/tags/v1.12.0.zip"
```
HTHou: I see the gtest has released v1.12.1 that also supports C++ 11, should we consider upgrading it?

hongzhi-gao: OK, I will upgrade the version to v1.12.1. Both v1.12.0 and v1.12.1 are stated to be the latest versions that support C++11.

HTHou (approved):
LGTM

PR #196, opened 2024-08-02
TsFile V4 for Table Model

jt2594838:


.github/workflows/unit-test.yml
```diff
@@ -77,6 +77,11 @@ jobs:
               core.setOutput('platform_suffix', ``)
             }
 
+      - name: Install python packages
+        if: ${{ matrix.os == 'ubuntu-latest'}}
+        run: pip3 install cython pandas numpy
+
+
```
HTHou: This part is useless anymore and can be removed.

cpp/pom.xml
```diff
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.apache.tsfile</groupId>
         <artifactId>tsfile-parent</artifactId>
-        <version>1.0.1-SNAPSHOT</version>
+        <version>1.0.1-tsfilev4-240801-SNAPSHOT</version>
```
HTHou: Should we keep the version 1.0.1-SNAPSHOT? We can change it later after the PR merged.

.github/workflows/unit-test.yml
jt2594838: Removed

cpp/pom.xml
jt2594838: This PR is just to trigger the CI, and I will not merge it into 'develop'.
Actually, I will merge it back to tsfile_v4.

jt2594838: How about we update it to 1.2.0-SNAPSHOT now?

JackieTien97 (approved):

HTHou (approved):

PR #200, opened 2024-08-05
Generating filters with freemarker templates

FearfulTomcat27:


java/tsfile/src/main/java/org/apache/tsfile/read/filter/basic/Filter.java
```diff
@@ -82,6 +82,14 @@ public abstract class Filter {
    */
   public abstract boolean[] satisfyTsBlock(TsBlock tsBlock);
```
JackieTien97, line 84: do we still need this method in IoTDB? if we just to keep compatibility, we can add @Deprecated annotation for this method.

java/tsfile/src/main/java/org/apache/tsfile/read/filter/basic/ValueFilter.java
```diff
@@ -75,6 +75,24 @@ public boolean[] satisfyTsBlock(TsBlock tsBlock) {
     return satisfyInfo;
   }
 
+  @Override
+  public boolean[] satisfyTsBlock(boolean[] selection, TsBlock tsBlock) {
```
JackieTien97, line 79: Will this method be used in new generated fller?

java/tsfile/src/main/java/org/apache/tsfile/utils/ReadWriteIOUtils.java
```diff
@@ -95,6 +95,12 @@ public static boolean readBool(ByteBuffer buffer) {
     return a == 1;
   }
 
+  /** read a bool from byteBuffer. */
+  public static boolean readBoolean(ByteBuffer buffer) {
```
JackieTien97, line 99: why not using existing `readBool`?

JackieTien97: Even if you want to keep the code same for each type in serialize method for generated code, you still have another two ways to achieve this goal:
1. call readBool in readBoolean
2. add if-else and specify readBool for boolean type in ftl.

JackieTien97 (changes requested):

JackieTien97:
BTW, you should add some description for your pr, including what you did in this pr and why you want to do this, and also link your performance test result.

PR #201, opened 2024-08-05
Fixed testing bug

CloudWise-Lukemiao:
Fixed testing bug.Mainly includes not outputting logs, outputting file directories without the original directory hierarchy, missing parameters without exception prompts, etc

### DIFF
--- a/java/tools/src/main/java/org/apache/tsfile/tools/TsFileTool.java
+++ b/java/tools/src/main/java/org/apache/tsfile/tools/TsFileTool.java
@@ -39,7 +39,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.BufferedReader;
+import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
@@ -97,6 +99,7 @@ public class TsFileTool {
       } catch (InterruptedException e) {
         LOGGER.error("Failed to await termination", e);
       }
+      LOGGER.info("The " + inputDirectoryStr + " directory or file has completed execution");
     }
   }
 
@@ -140,8 +143,14 @@ public class TsFileTool {
     return new TableSchema(tableName, measurementSchemas, columnTypes);
   }
 
-  private static boolean writeTsFile(String fileName, List<String> lineList) {
-    final File tsFile = new File(outputDirectoryStr, fileName);
+  private static boolean writeTsFile(
+      String sourceFilePath, String fileName, List<String> lineList) {
+    String inputFileAbsolutePath = new File(inputDirectoryStr).getAbsolutePath();
+    String soureFlieName = new File(sourceFilePath).getName();
+    String fileOutPutDirStr =
+        outputDirectoryStr
+            + sourceFilePath.replace(inputFileAbsolutePath, "").replace(soureFlieName, "");
+    final File tsFile = new File(fileOutPutDirStr, fileName);
     TsFileWriter writer = null;
     try {
       writer = new TsFileWriter(tsFile);
@@ -158,17 +167,22 @@ public class TsFileTool {
         return false;
       }
     } catch (Exception e) {
-      e.printStackTrace();
-      LOGGER.error("Failed to write file: " + tsFile);
+      LOGGER.error("Failed to write file" + tsFile.getAbsolutePath() + ":{} ", e.getMessage());
       return false;
     } finally {
       if (writer != null) {
         try {
           writer.close();
         } catch (IOException e) {
-          e.printStackTrace();
+          LOGGER.error("Failed to close file" + tsFile.getAbsolutePath() + ":{}", e.getMessage());
         }
       }
+    }
+  }
+
+  private static void deleteFile(File tsfile) {
+    if (!tsfile.delete()) {
+      LOGGER.error(tsfile.getAbsolutePath() + " delete failed");
     }
   }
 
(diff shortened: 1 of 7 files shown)


CritasWang (approved):
LGTM

java/tools/src/main/java/org/apache/tsfile/tools/TsFileTool.java
```diff
@@ -158,20 +167,25 @@ private static boolean writeTsFile(String fileName, List<String> lineList) {
...
-      LOGGER.error("Failed to write file: " + tsFile);
+      LOGGER.error("Failed to write file" + tsFile.getAbsolutePath() + ":{} ", e.getMessage());
       return false;
     } finally {
       if (writer != null) {
         try {
           writer.close();
         } catch (IOException e) {
-          e.printStackTrace();
+          LOGGER.error("Failed to close file" + tsFile.getAbsolutePath() + ":{}", e.getMessage());
```
jt2594838: Are you sure the exception stack is unnecessary? If not, may use
 LOGGER.error("Failed to close file: " + tsFile.getAbsolutePath(), e);

java/tools/src/main/java/org/apache/tsfile/tools/TsFileTool.java
```diff
@@ -188,6 +202,7 @@ private static Tablet genTablet(
       map.put(column.name, i);
     }
     try {
+      sortByColumnIndex(lineList);
       for (int i = 0; i < num; i++) {
         String line = lineList.get(i);
         String[] lineArray = line.split(schema.separator);
```
jt2594838: The lines are split twice, which is unnecessary.
You may first split them, sort them, and then iterate over them.

java/tools/src/main/java/org/apache/tsfile/tools/TsFileTool.java
```diff
@@ -266,15 +296,57 @@ private static void processDirectory(File directory, ExecutorService executor) {
...
+              new InputStreamReader(
+                  Files.newInputStream(Paths.get(fileAbsolutePath)), StandardCharsets.UTF_8))) {
+        String line;
+        if ((line = reader.readLine()) != null) {
+          data.add(0, line);
+        }
+      } catch (Exception e) {
+        LOGGER.error("Error reading file: " + e.getMessage());
+      }
+    }
```
jt2594838: First, how about caching the header line as a member field in the previous process.
Second, I would assume the size of data to be quite large, so inserting an element to the head may not be efficient if using an ArrayList, and a LinkedList is also inefficient due to the list item overhead. You may just store the header line in a local variable to avoid this.

java/tools/src/main/java/org/apache/tsfile/tools/TsFileTool.java
jt2594838: Please also check other places.

java/tools/src/test/java/org/apache/tsfile/tools/TsfiletoolsTest.java
jt2594838, line 1: The new version seems to be able to move erroneous input files to the target directory. How about adding associated test cases?

jt2594838 (approved):
Good job, but with some minor issues.

java/tools/src/test/java/org/apache/tsfile/tools/TsfiletoolsTest.java
```diff
@@ -95,6 +103,35 @@ public void genSchemaFile() {
...
+          writer.write(
+              timestamp + "aa" + "," + tmp1 + "," + tmp2 + "," + tmp3 + "," + tmp4 + "," + tmp5);
+        } else {
+          writer.write(timestamp + "," + tmp1 + "," + tmp2 + "," + tmp3 + "," + tmp4 + "," + tmp5);
+        }
+
+        writer.newLine();
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
```
jt2594838: May use `fail(e.getMessage())` or simply throw the exception, since this is a test.

java/tools/src/test/java/org/apache/tsfile/tools/TsfiletoolsTest.java
```diff
@@ -162,4 +199,31 @@ public void testCsvToTsfile() throws Exception {
...
+    String scFilePath = new File(schemaFile).getAbsolutePath();
+    String csvFilePath = new File(wrongCsvFile).getAbsolutePath();
+    String targetPath = new File(testDir).getAbsolutePath();
+    String fd = new File(failedDir).getAbsolutePath();
+    String[] args =
+        new String[] {
+          "-s" + csvFilePath, "-schema" + scFilePath, "-t" + targetPath, "-fail_dir" + fd
+        };
+    TsFileTool.main(args);
+    if (new File(failedDir + File.separator + "dataWrong.csv").exists()) {
```
jt2594838: assertTrue(new File(failedDir + File.separator + "dataWrong.csv").exists())

java/tools/src/test/java/org/apache/tsfile/tools/TsfiletoolsTest.java
```diff
@@ -162,4 +199,31 @@ public void testCsvToTsfile() throws Exception {
...
+      try (BufferedReader br = new BufferedReader(new FileReader(wrongCsvFile))) {
+        int num = 0;
+        while (br.readLine() != null) {
+          num++;
+        }
+        assertEquals(101, num);
+      } catch (IOException e) {
+        e.printStackTrace();
+        Assert.assertTrue(false);
```
jt2594838: fail(e.getMessage()) or simply throw the exception.

jt2594838 (changes requested):
Please take a look at the failed tests.

jt2594838 (approved):

PR #214, opened 2024-08-25
Enable address sanitizer in tsfile cpp.

ColinLeeo:
In this pr:
1. Fixed the issue where a flush operation would result in an E_ALLREAY_EXIST error after inserting a large amount of data.
2. Enabled Address Sanitizer in tests.
3. Checked and fixed memory leaks.
4. Removed some unused configuration items.
5. Performed code formatting and cleanup.

### DIFF
--- a/cpp/examples/cpp_examples/demo_write.cpp
+++ b/cpp/examples/cpp_examples/demo_write.cpp
@@ -51,15 +51,15 @@ static std::string generate_random_string(int length) {
 int demo_write() {
     TsFileWriter* tsfile_writer_ = new TsFileWriter();
     libtsfile_init();
-    std::cout<<"input tablesize"<<std::endl;
+    std::cout << "input tablesize" << std::endl;
     std::string file_name_ = std::string("tsfile_writer_test_") +
                              generate_random_string(10) +
                              std::string(".tsfile");
     int flags = O_WRONLY | O_CREAT | O_TRUNC;
 #ifdef _WIN32
-        flags |= O_BINARY;
+    flags |= O_BINARY;
 #endif
-        mode_t mode = 0666;
+    mode_t mode = 0666;
     tsfile_writer_->open(file_name_, flags, mode);
     remove(file_name_.c_str());
     const int device_num = 50;
@@ -83,7 +83,7 @@ int demo_write() {
     int max_rows = 100000;
     int table_size;
     std::cin >> table_size;
-    std::cout<<"start"<<std::endl;
+    std::cout << "start" << std::endl;
     int cur_row = 0;
     long start = getNowTime();
     for (; cur_row < max_rows;) {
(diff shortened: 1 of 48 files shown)


cpp/src/common/tsfile_common.h
```diff
@@ -580,6 +581,13 @@ class TSMIterator {
     common::SimpleList<ChunkGroupMeta *> &chunk_group_meta_list_;
     common::SimpleList<ChunkGroupMeta *>::Iterator chunk_group_meta_iter_;
     common::SimpleList<ChunkMeta *>::Iterator chunk_meta_iter_;
+    std::map<common::String, std::map<common::String, std::vector<ChunkMeta *>>>
+        tsm_info_;
+    std::map<common::String,
+             std::map<common::String, std::vector<ChunkMeta *>>>::iterator
+        tsm_device_iter_;
+    std::map<common::String, std::vector<ChunkMeta *>>::iterator
+        tsm_measurement_iter_;
```
jt2594838: Add some comments to explain what each String indicates or use better names.

jt2594838 (approved):

ColinLeeo:
It seems that compiling tsfile_cpp with asan at Windows needs clang instead of Mingw.

jt2594838:
> It seems that compiling tsfile_cpp with asan at Windows needs clang instead of Mingw.

Yeah, how about only enabling it on linux. I think it enough for detecting memory leaks.

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -513,6 +513,8 @@ int AlignedChunkReader::decode_time_value_buf_into_tsblock(
         if (!prev_time_page_not_finish()) {
             time_in_.reset();
         }
+        value_page_col_notnull_bitmap_.clear();
```
ColinLeeo, line 516: Frequent resizing of vectors can lead to high memory usage and may cause AddressSanitizer (ASan) to report errors.

cpp/test/file/open_file_test.cc
```diff
@@ -40,11 +40,6 @@ class OpenFileTest : public ::testing::Test {
     OpenFile* open_file_;
 };
 
-TEST_F(OpenFileTest, Initialization) {
```
ColinLeeo, line 43: This unit test is unnecessary and can cause memory leaks during testing.

cpp/src/common/datatype/value.h
```diff
@@ -33,16 +33,41 @@
...
+                break;
+            case TEXT:
+                value_.sval_ = nullptr;
+                break;
+            case NULL_TYPE:
+                break;
+            default:
+                LOGE("unknown data type");
+        }
+    }
```
ColinLeeo, line 61: Initializing a union by setting only a single variable to 0 may not be sufficient, as the variable might not be long enough, causing other types to remain non-zero.

cpp/examples/cpp_examples/demo_write.cpp
```diff
@@ -20,39 +20,97 @@
...
-    int row_count = 100;
-    for (int i = 1; i < row_count; ++i) {
-        storage::DataPoint point(measurement_name, 10000 + i);
-        storage::TsRecord record(i, device_name, 1);
-        record.points_.push_back(point);
-        ret = tsfile_writer.write_record(record);
-        ASSERT(ret == 0);
+    TsFileWriter* tsfile_writer_ = new TsFileWriter();
+    libtsfile_init();
+    std::cout << "input tablesize" << std::endl;
```
jt2594838: This line should be moved to the line before line 85, namely 'std::cin >> table_size;'.
And beware of the terms "Table" and "Tablet".

jt2594838 (approved):

jt2594838 (approved):

PR #222, opened 2024-08-30
Fix TsFileDeviceIterator

shuwenwei:
Fix TsFileDeviceIterator

### DIFF
--- a/java/tsfile/src/test/java/org/apache/tsfile/read/TsFileDeviceIteratorTest.java
+++ b/java/tsfile/src/test/java/org/apache/tsfile/read/TsFileDeviceIteratorTest.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


java/tsfile/src/test/java/org/apache/tsfile/read/TsFileDeviceIteratorTest.java
```diff
@@ -0,0 +1,136 @@
...
+      IDeviceID previous = null;
+      while (deviceIterator.hasNext()) {
+        Pair<IDeviceID, Boolean> next = deviceIterator.next();
+        if (previous != null) {
+          Assert.assertTrue(previous.compareTo(next.getLeft()) < 0);
+        }
+        previous = next.getLeft();
+      }
+    }
+  }
```
jt2594838: The number of returned devices should be asserted in case deviceIterator does not return anything.

jt2594838 (approved):

java/tsfile/src/test/java/org/apache/tsfile/read/TsFileDeviceIteratorTest.java
shuwenwei: added

jt2594838 (approved):

PR #233, opened 2024-09-03
Use reflection method to generate encrypt instances

zhujt20:
Use reflection method instead of direct ways to generate encrypt instances.


java/tsfile/src/main/java/org/apache/tsfile/encrypt/AES128Decryptor.java
```diff
@@ -0,0 +1,85 @@
...
+    try {
+      return AES.doFinal(data);
+    } catch (IllegalBlockSizeException | BadPaddingException e) {
+      throw new EncryptException("AES128Decryptor decrypt failed ", e);
+    }
+  }
+
+  @Override
+  public byte[] decrypt(byte[] data, int offset, int size) {
+    //    return decrypt(Arrays.copyOfRange(data, offset, offset + size));
```
jt2594838: Remove

java/tsfile/src/main/java/org/apache/tsfile/encrypt/AES128Encryptor.java
```diff
@@ -0,0 +1,86 @@
...
+        | InvalidKeyException e) {
+      throw new EncryptException("AES128Encryptor init failed ", e);
+    }
+  }
+
+  @Override
+  public byte[] encrypt(byte[] data) {
+    try {
+      return AES.doFinal(data);
+      //      return AES.doFinal(data);
```
jt2594838: Remove, check other places.

java/tsfile/src/main/java/org/apache/tsfile/encrypt/EncryptUtils.java
```diff
@@ -109,37 +158,61 @@ public static IEncryptor getDefaultEncryptor() {
...
-      encryptType = EncryptionType.UNENCRYPTED;
+      encryptType = "org.apache.tsfile.encrypt.UNENCRYPTED";
       dataEncryptKey = null;
     }
-    return IEncryptor.getEncryptor(encryptType, dataEncryptKey);
+    try {
+      Class<?> encryptTypeClass = Class.forName(encryptType);
+      java.lang.reflect.Constructor<?> constructor =
+          encryptTypeClass.getDeclaredConstructor(byte[].class);
+      return ((IEncrypt) constructor.newInstance(dataEncryptKey));
```
jt2594838, line 173: The constructor may be stored as a field to reduce the number of reflection calls.

java/tsfile/src/main/java/org/apache/tsfile/encrypt/IDecryptor.java
```diff
@@ -20,47 +20,30 @@
...
-        return new AES128Decryptor(key);
-      default:
-        logger.warn("Unknown encryption type: {}", name);
-        return new NoDecryptor();
+  static IDecryptor getDecryptor(String type, byte[] key) {
+    try {
+      Class<?> encryptClass = Class.forName(type);
+      java.lang.reflect.Constructor<?> constructor =
+          encryptClass.getDeclaredConstructor(byte[].class);
+      return ((IEncrypt) constructor.newInstance(key)).getDecryptor();
```
jt2594838: May try using a concurrent map to store the constructor.

java/tsfile/src/main/java/org/apache/tsfile/read/TsFileSequenceReader.java
```diff
@@ -1721,8 +1721,7 @@ private static ByteBuffer decrypt(IDecryptor decryptor, ByteBuffer buffer) {
       return buffer;
     }
     return ByteBuffer.wrap(
-        decryptor.decrypt(
-            buffer.array(), buffer.arrayOffset() + buffer.position(), buffer.remaining()));
+        decryptor.decrypt(buffer.array(), buffer.position(), buffer.remaining()));
```
jt2594838: Why is buffer.arrayOffset() removed?

jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/encrypt/EncryptUtils.java
zhujt20: Actually the constructor here will only be used once.

java/tsfile/src/main/java/org/apache/tsfile/encrypt/IDecryptor.java
zhujt20: For decrypt, there may be different kinds of constructors to be used in the future. So maybe storing them is a good idea, but where should I implement this?

jt2594838: Just like the logger above, add a static member.

jt2594838 (approved):

PR #259, opened 2024-10-08
Add isTableModel method to IDeviceID 

shuwenwei:


jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/file/metadata/StringArrayDeviceID.java
```diff
@@ -211,6 +212,11 @@ public boolean isEmpty() {
     return segments == null || segments.length == 0;
   }
 
+  @Override
+  public boolean isTableModel() {
+    return segments[0].startsWith(PATH_ROOT + PATH_SEPARATOR);
+  }
```
jt2594838: The condition should be reversed. May add a small test.

jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/file/metadata/StringArrayDeviceID.java
shuwenwei: fixed

jt2594838 (approved):

PR #278, opened 2024-10-29
encrypt thread safety

zhujt20:
Push down the encryptor instance build time for thread safety.

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/encrypt/EncryptUtils.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/encrypt/EncryptUtils.java
@@ -39,11 +39,7 @@ public class EncryptUtils {
 
   public static String normalKeyStr = getNormalKeyStr();
 
-  public static IEncrypt encrypt = getEncrypt();
-
-  public static IEncryptor encryptor = encrypt.getEncryptor();
-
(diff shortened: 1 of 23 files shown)


java/tsfile/src/main/java/org/apache/tsfile/encrypt/EncryptUtils.java
```diff
@@ -148,6 +144,48 @@ public static String getNormalKeyStr(TSFileConfig conf) {
...
+        dataEncryptKey = md.digest();
+      } catch (Exception e) {
+        throw new EncryptException(
+            "md5 function not found while using md5 to generate data key", e);
+      }
+    } else {
+      encryptType = "org.apache.tsfile.encrypt.UNENCRYPTED";
+      dataEncryptKey = null;
+    }
+    return new EncryptParameter(encryptType, dataEncryptKey);
```
jt2594838: May simplify with `return getEncryptParameter(TSFileDescriptor.getInstance().getConfig())` to reduce duplicated code.

jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/encrypt/EncryptUtils.java
zhujt20: ok，already simplified.

PR #284, opened 2024-11-04
Fix the problem that ChunkMeta of empty Chunk is not processed

luoluoyuyu:
As the title said

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/read/reader/series/FileSeriesReader.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/read/reader/series/FileSeriesReader.java
@@ -57,7 +57,6 @@ public class FileSeriesReader extends AbstractFileSeriesReader {
       List<Chunk> valueChunkList = new ArrayList<>();
       for (IChunkMetadata metadata : alignedChunkMetadata.getValueChunkMetadataList()) {
         if (metadata != null) {
-          System.out.println(metadata);
           valueChunkList.add(chunkLoader.loadChunk((ChunkMetadata) metadata));
           currentChunkMeasurementNames.add(metadata.getMeasurementUid());
           continue;
--- a/java/tsfile/src/main/java/org/apache/tsfile/read/reader/series/FileSeriesReader.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/read/reader/series/FileSeriesReader.java
@@ -56,8 +56,10 @@ public class FileSeriesReader extends AbstractFileSeriesReader {
           chunkLoader.loadChunk((ChunkMetadata) (alignedChunkMetadata.getTimeChunkMetadata()));
       List<Chunk> valueChunkList = new ArrayList<>();
       for (IChunkMetadata metadata : alignedChunkMetadata.getValueChunkMetadataList()) {
-        valueChunkList.add(chunkLoader.loadChunk((ChunkMetadata) metadata));
-        currentChunkMeasurementNames.add(metadata.getMeasurementUid());
+        if (metadata != null) {
+          valueChunkList.add(chunkLoader.loadChunk((ChunkMetadata) metadata));
+          currentChunkMeasurementNames.add(metadata.getMeasurementUid());
+        }
       }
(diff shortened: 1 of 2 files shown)


jt2594838 (approved):
LGTM, but I am not sure if it should be skipped or a place holder is needed.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/series/FileSeriesReader.java
```diff
@@ -56,8 +56,10 @@ protected void initChunkReader(IChunkMetadata chunkMetaData) throws IOException
           chunkLoader.loadChunk((ChunkMetadata) (alignedChunkMetadata.getTimeChunkMetadata()));
       List<Chunk> valueChunkList = new ArrayList<>();
       for (IChunkMetadata metadata : alignedChunkMetadata.getValueChunkMetadataList()) {
-        valueChunkList.add(chunkLoader.loadChunk((ChunkMetadata) metadata));
-        currentChunkMeasurementNames.add(metadata.getMeasurementUid());
+        if (metadata != null) {
```
shuwenwei, line 59: When the value chunk metadata is null, is it necessary to ensure that the valueChunkList in AlignedChunkReader aligns with the original AlignedChunkMetadata at every value column?

shuwenwei: If there are value columns (s1, s2, s3), in the first AlignedChunkMetadata, s2 corresponds to metadata as null, at this time the returned valueChunkList is (s1, s3). In the subsequent AlignedChunkMetadata, s3 corresponds to metadata as null, at this time the returned valueChunkList is (s1, s2). Perhaps the two returns should be (s1, null, s3) (s1, s2, null)

luoluoyuyu: I think it is unnecessary, because if Chunk is null, no subsequent read operations can be performed, and if ChunkMeta is null or ChunkMeta count is 0, it is saved as null in ChunkMetaList.

luoluoyuyu: But I have a question, does currentChunkMeasurementNames need to correspond to Chunk? If valueChunkList  is (s1,s2,null), does currentChunkMeasurementNames also need to correspond to (s1,s2,null)?

luoluoyuyu: > If there are value columns (s1, s2, s3), in the first AlignedChunkMetadata, s2 corresponds to metadata as null, at this time the returned valueChunkList is (s1, s3). In the subsequent AlignedChunkMetadata, s3 corresponds to metadata as null, at this time the returned valueChunkList is (s1, s2). Perhaps the two returns should be (s1, null, s3) (s1, s2, null)

Does `currentChunkMeasurementNames` already solve this problem well?

shuwenwei: > > If there are value columns (s1, s2, s3), in the first AlignedChunkMetadata, s2 corresponds to metadata as null, at this time the returned valueChunkList is (s1, s3). In the subsequent AlignedChunkMetadata, s3 corresponds to metadata as null, at this time the returned valueChunkList is (s1, s2). Perhaps the two returns should be (s1, null, s3) (s1, s2, null)
> 
> Does `currentChunkMeasurementNames` already solve this problem well?

Yes, I checked the usage of this class, and in SingleDeviceTsBlockReader, the issue is resolved by using getCurrentChunkMeasurementNames method. Another usage of this class is in TsFileReader, where the QueryExpression is conducted based on the input Path. It seems that even if each Path is from the same aligned series of value columns, it still proceeds in multiple FileSeriesReader instances, hence the AlignedChunkMetadata here may only have one value column's chunk metadata, so it also seems to be correct.
截屏2024-11-06 14 22 50

jt2594838 (approved):

PR #295, opened 2024-11-07
Tsfile java interfaces

shuwenwei:


java/tsfile/src/main/java/org/apache/tsfile/common/TsFileApi.java
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.tsfile.common;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.SOURCE)
```
jt2594838: Better to use CLASS retention so that it is still visible even if only a Jar is used.

jt2594838 (approved):

PR #308, opened 2024-11-20
Add PooledBinary Class

MrQuansy:
# Summary
Add PooledBinary class for Binary object allocation optimization.

This PR introduces a new PooledBinary class designed for binary object pooling at the application layer. It aims to improve allocation performance and reduce GC overhead by enabling binary object reuse.

# Details
The PooledBinary class is similar to the existing Binary class but with additional features to support object pooling:

* Added length field to indicate the actual valid data length, which is less than or equal to the underlying byte array length
* Added arenaIndex field to serve the allocator's needs

# Next Steps
This is the first step in a series of planned changes. Future PRs will modify specific Binary usages to PooledBinary in the codebase.

### DIFF
--- a/java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
+++ b/java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
@@ -18,8 +18,6 @@
  */
 package org.apache.tsfile.utils;
 
-import jdk.internal.util.ArraysSupport;
-
 import java.io.Serializable;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
@@ -105,8 +103,16 @@ public class PooledBinary implements Comparable<PooledBinary>, Serializable, Acc
       return false;
     }
 
-    return ArraysSupport.mismatch(binary.getValues(), 0, otherBinary.binary.getValues(), 0, length)
-        == -1;
+    byte[] v0 = binary.getValues();
+    byte[] v1 = otherBinary.binary.getValues();
+
+    for (int i = 0; i < length; i++) {
+      if (v0[i] != v1[i]) {
+        return false;
+      }
+    }
+
+    return true;
   }
 
   @Override
--- a/java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
+++ b/java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
@@ -24,6 +24,7 @@ import java.nio.charset.StandardCharsets;
 
 import static org.apache.tsfile.utils.RamUsageEstimator.shallowSizeOfInstance;
 import static org.apache.tsfile.utils.RamUsageEstimator.sizeOf;
+
 /**
  * This class represents a pooled binary object for application layer. It is designed to improve
  * allocation performance and reduce GC overhead by reusing binary objects from a pool instead of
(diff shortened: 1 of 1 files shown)


java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
```diff
@@ -0,0 +1,176 @@
...
+ * length of the underlying byte array. Always use getLength() instead of getValue().length to get
+ * the correct length.
+ */
+public class PooledBinary implements Comparable<PooledBinary>, Serializable, Accountable {
+
+  private static final long INSTANCE_SIZE = shallowSizeOfInstance(PooledBinary.class);
+  private static final long serialVersionUID = 6394197743397020735L;
+  public static final PooledBinary EMPTY_VALUE = new PooledBinary(new byte[0]);
+
+  private Binary values;
```
OneSizeFitsQuorum: rename to `binary`?

java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    PooledBinary binary = (PooledBinary) o;
+
+    if (length != binary.getLength()) {
+      return false;
+    }
+    for (int i = 0; i < length; i++) {
```
OneSizeFitsQuorum, line 109: Please use this API
image

Because it's vectorized.
image

java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
```diff
@@ -0,0 +1,176 @@
...
+    return true;
+  }
+
+  @Override
+  public int hashCode() {
+    if (values.getValues() == null) {
+      return 0;
+    } else {
+      int result = 1;
+      byte[] var2 = values.getValues();
```
OneSizeFitsQuorum: can you make them meaningful? Do not use var1 var2 var3

java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+  }
+
+  /**
+   * get length.
+   *
+   * @return length
+   */
+  public int getLength() {
+    return this.length;
```
OneSizeFitsQuorum, line 136: if the value of binary is null, we should return -1?

OneSizeFitsQuorum (commented):
PTAL

java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
MrQuansy: The binary class already contains this logic.

java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
MrQuansy: Make sense

OneSizeFitsQuorum (approved):
LGTM

java/common/src/main/java/org/apache/tsfile/utils/PooledBinary.java
OneSizeFitsQuorum: not compatible with jdk8 , please ignore this comment

OneSizeFitsQuorum (approved):
LGTM

PR #327, opened 2024-12-09
Add TableModel annotations

shuwenwei:
Add TableModel annotations

### DIFF
--- a/java/common/src/main/java/org/apache/tsfile/annotations/TableModel.java
+++ b/java/common/src/main/java/org/apache/tsfile/annotations/TableModel.java
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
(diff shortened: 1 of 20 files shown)


java/common/src/main/java/org/apache/tsfile/annotations/TableModel.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation indicates that the related filed, method, or constructor is only used for
+ * TableModel.
+ */
+@Target({ElementType.METHOD, ElementType.FIELD, ElementType.CONSTRUCTOR})
+@Retention(RetentionPolicy.SOURCE)
```
jt2594838: Should set it to class so that people who are just using maven dependency can also seem them.

jt2594838 (approved):

PR #328, opened 2024-12-09
Add FlushChunkMetadataListener

DanielWang2035:
When all ChunkGroupMetadata and ChunkMetadata are flushed into temporary files, notify all registered listeners. This will be used in loading tsfile.

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/write/writer/FlushChunkMetadataListener.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/write/writer/FlushChunkMetadataListener.java
@@ -20,7 +20,7 @@
 package org.apache.tsfile.write.writer;
 
 import org.apache.tsfile.file.metadata.IChunkMetadata;
-import org.apache.tsfile.read.common.Path;
+import org.apache.tsfile.file.metadata.IDeviceID;
 import org.apache.tsfile.utils.Pair;
 
 import java.util.List;
@@ -28,6 +28,6 @@ import java.util.List;
 @FunctionalInterface
 public interface FlushChunkMetadataListener {
 
-  // measurement id -> chunk metadata list
-  void onFlush(List<Pair<Path, List<IChunkMetadata>>> sortedChunkMetadataList);
+  // Pair<device id, measurement id> -> chunk metadata list
+  void onFlush(List<Pair<Pair<IDeviceID, String>, List<IChunkMetadata>>> sortedChunkMetadataList);
 }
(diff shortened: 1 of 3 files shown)


SteveYurongSu (approved):

java/tsfile/src/main/java/org/apache/tsfile/write/writer/FlushChunkMetadataListener.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.tsfile.read.common.Path;
+import org.apache.tsfile.utils.Pair;
+
+import java.util.List;
+
+@FunctionalInterface
+public interface FlushChunkMetadataListener {
+
+  // measurement id -> chunk metadata list
+  void onFlush(List<Pair<Path, List<IChunkMetadata>>> sortedChunkMetadataList);
```
jt2594838: Is this method also used by the TableModel? If so, may avoid using Path and use Pair<IDeviceId, String> instead.
(That is, do not expose Path to TableModel users.)

java/tsfile/src/main/java/org/apache/tsfile/write/writer/TsFileIOWriter.java
```diff
@@ -131,6 +132,8 @@ public class TsFileIOWriter implements AutoCloseable {
 
   protected String encryptKey;
 
+  private final List<FlushChunkMetadataListener> flushListeners = new CopyOnWriteArrayList<>();
```
jt2594838: Is it necessary to use copy-on-write? The operations on the same TsFile are mostly single-threaded.
In what scenario can `addFlushListener` and `flush()` be called concurrently?

java/tsfile/src/main/java/org/apache/tsfile/write/writer/TsFileIOWriter.java
jt2594838, line 1: Better to add some tests.

jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/write/writer/FlushChunkMetadataListener.java
DanielWang2035: fixed

java/tsfile/src/main/java/org/apache/tsfile/write/writer/TsFileIOWriter.java
DanielWang2035: This is just in case. fixed

java/tsfile/src/main/java/org/apache/tsfile/write/writer/TsFileIOWriter.java
DanielWang2035: fixed

jt2594838 (approved):

PR #370, opened 2025-01-08
Implement DictionaryColumn and support copyPositions and  getPositions method in Column

Wei-hao-Li:
Add `copyPositions` because: DictionaryColumn need to implement `getRegionCopy` method, it calls `compact` method of  
 DictionaryColumn, which calls `copyPositions`.

### DIFF
--- a/java/common/src/main/java/org/apache/tsfile/block/column/Column.java
+++ b/java/common/src/main/java/org/apache/tsfile/block/column/Column.java
@@ -170,6 +170,23 @@ public interface Column {
   /** This method will create a copy of origin column with different array offset. */
   Column subColumnCopy(int fromIndex);
 
+  /**
+   * Create a new colum from the current colum by keeping the same elements only with respect to
+   * {@code positions} that starts at {@code offset} and has length of {@code length}. The
+   * implementation may return a view over the data in this colum or may return a copy, and the
+   * implementation is allowed to retain the positions array for use in the view.
+   */
+  Column getPositions(int[] positions, int offset, int length);
+
+  /**
+   * Returns a column containing the specified positions. Positions to copy are stored in a subarray
+   * within {@code positions} array that starts at {@code offset} and has length of {@code length}.
+   * All specified positions must be valid for this block.
+   *
+   * <p>The returned column must be a compact representation of the original column.
+   */
+  Column copyPositions(int[] positions, int offset, int length);
+
   /** reverse the column */
   void reverse();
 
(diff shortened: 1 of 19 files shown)


java/common/src/main/java/org/apache/tsfile/block/column/Column.java
```diff
@@ -170,6 +170,23 @@ default TsPrimitiveType getTsPrimitiveType(int position) {
   /** This method will create a copy of origin column with different array offset. */
   Column subColumnCopy(int fromIndex);
 
+  /**
+   * Create a new colum from the current colum by keeping the same elements only with respect to
+   * {@code positions} that starts at {@code offset} and has length of {@code length}. The
+   * implementation may return a view over the data in this colum or may return a copy, and the
+   * implementation is allowed to retain the positions array for use in the view.
+   */
+  Column getPositions(int[] positions, int offset, int length);
```
JackieTien97, line 179: why not use default method here?

java/tsfile/src/main/java/org/apache/tsfile/read/common/block/column/RunLengthEncodedColumn.java
```diff
@@ -221,6 +223,25 @@ public Column subColumnCopy(int fromIndex) {
     return new RunLengthEncodedColumn(valueCopy, positionCount - fromIndex);
   }
 
+  @Override
+  public Column getPositions(int[] positions, int offset, int length) {
+    checkArrayRange(positions, offset, length);
+
+    return DictionaryColumn.createInternal(
+        offset, length, this, positions, DictionaryId.randomDictionaryId());
```
JackieTien97: why we still need `DictionaryColumn` for RLEColumn?

java/tsfile/src/main/java/org/apache/tsfile/read/common/block/column/RunLengthEncodedColumn.java
```diff
@@ -221,6 +223,25 @@ public Column subColumnCopy(int fromIndex) {
...
+
+  @Override
+  public Column copyPositions(int[] positions, int offset, int length) {
+    checkArrayRange(positions, offset, length);
+
+    for (int position : positions) {
+      checkReadablePosition(this, position);
+    }
+
+    return new RunLengthEncodedColumn(value, length);
```
JackieTien97: do the copy thing?

JackieTien97 (changes requested):

JackieTien97:
throw exception for all getArrays and reset.

Also add another Encoding implementation for DictionaryColumn, remember to add ut in TsBlockSerdeTest

JackieTien97 (approved):

PR #382, opened 2025-01-15
Add statistic compatibility

jt2594838:


java/tsfile/src/main/java/org/apache/tsfile/file/metadata/statistics/IntegerStatistics.java
```diff
@@ -142,17 +141,17 @@ public Integer getLastValue() {
...
   @Override
   public long getSumLongValue() {
     return sumValue;
   }
 
+  @SuppressWarnings("rawtypes")
   @Override
-  protected void mergeStatisticsValue(Statistics<Integer> stats) {
+  protected void mergeStatisticsValue(Statistics stats) {
     IntegerStatistics intStats = (IntegerStatistics) stats;
```
shuwenwei: Why didn't modify IntegerStatistics?

jt2594838 (commented):
Check

java/tsfile/src/main/java/org/apache/tsfile/file/metadata/statistics/IntegerStatistics.java
jt2594838: No other type can be altered to INT32 so other statistics will not be merged into IntegerStatistics.

shuwenwei: how about throw an StatisticsClassException like the BooleanStatistics

jt2594838: Added

shuwenwei (approved):

PR #386, opened 2025-01-23
Don't print exception log when thread is interrupted

shuwenwei:


java/tsfile/src/main/java/org/apache/tsfile/write/writer/tsmiterator/DiskTSMIterator.java
```diff
@@ -78,22 +78,24 @@ public boolean hasNext() {
...
         return super.next();
       }
     } catch (IOException e) {
-      LOG.error("Meets IOException when reading timeseries metadata from disk", e);
-      return null;
+      if (!Thread.currentThread().isInterrupted()) {
+        LOG.error("Meets IOException when reading timeseries metadata from disk", e);
+      }
+      throw e;
     }
```
jt2594838, line 94: May catch ClosedByInterruptException and ignore it.

jt2594838 (approved):

jt2594838 (approved):

PR #409, opened 2025-02-14
C++ Tsfile V4 read

zwhzzz0821:
as the title mentioned.


cpp/src/common/cache/lru_cache.h
```diff
@@ -0,0 +1,226 @@
...
+    typedef KeyValuePair<Key, Value> node_type;
+    typedef std::list<KeyValuePair<Key, Value>> list_type;
+    typedef Map map_type;
+    typedef Lock lock_type;
+    using Guard = std::lock_guard<lock_type>;
+    /**
+     * the maxSize is the soft limit of keys and (maxSize + elasticity) is the
+     * hard limit
+     * the cache is allowed to grow till (maxSize + elasticity) and is pruned
+     * back to maxSize keys set maxSize = 0 for an unbounded cache (but in that
```
jt2594838: "back to maxSize keys set maxSize = 0"
add a full stop between “maxSize keys” and "set" or start a new line.

cpp/src/common/cache/lru_cache.h
```diff
@@ -0,0 +1,226 @@
...
+        keys_.emplace_front(k, std::move(v));
+        cache_[k] = keys_.begin();
+        prune();
+    }
+    void emplace(const Key& k, Value&& v) {
+        Guard g(lock_);
+        keys_.emplace_front(k, std::move(v));
+        cache_[k] = keys_.begin();
+        prune();
+    }
```
jt2594838: Is this method necessary? If `emplace` is called for the same key twice, or mixed with `insert,` then the integrity of the cache will be compromised.

cpp/src/common/cache/lru_cache.h
```diff
@@ -0,0 +1,226 @@
...
+    bool tryGetRef_nolock(const Key& kIn, Value& vOut) {
+        const auto iter = cache_.find(kIn);
+        if (iter == cache_.end()) {
+            return false;
+        }
+        keys_.splice(keys_.begin(), keys_, iter->second);
+        vOut = iter->second->value;
+        return true;
+    }
```
jt2594838: The two methods seem highly alike. Are both of them necessary?

cpp/src/common/cache/lru_cache.h
```diff
@@ -0,0 +1,226 @@
...
+
+   private:
+    // Disallow copying.
+    Cache(const Cache&) = delete;
+    Cache& operator=(const Cache&) = delete;
+
+    mutable Lock lock_;
+    Map cache_;
+    list_type keys_;
```
jt2594838: keys_ -> entries_

cpp/src/common/schema.h
```diff
@@ -20,319 +20,342 @@
...
+        auto it = column_pos_index_.find(lower_case_column_name);
+        if (it != column_pos_index_.end()) {
+            return it->second;
+        } else {
+            int index = -1;
+            for (size_t i = 0; i < column_schemas_.size(); ++i) {
+                if (to_lower(column_schemas_[i]->measurement_name_) ==
+                    lower_case_column_name) {
+                    index = static_cast<int>(i);
+                    break;
```
jt2594838: measurement names should be converted to lower cases when put into the schema, then it is necessary to call to_lower here.

cpp/src/common/schema.h
```diff
@@ -20,319 +20,342 @@
...
+    }
+
+    common::ColumnDesc get_column_desc(const std::string &column_name) {
+        int column_idx = find_column_index(column_name);
+        return common::ColumnDesc(
+            column_schemas_[column_idx]->data_type_,
+            column_schemas_[column_idx]->encoding_,
+            column_schemas_[column_idx]->compression_type_, INVALID_TTL,
+            column_name, common::TsID());
+    }
```
jt2594838: ColumnDesc is no longer feasible for the current TsFile.
There is no TTL for TsFile and TsID is deprecated.
Should use ColumnSchema instead, referring to https://timechor.feishu.cn/docx/WszvdVt1polnbIxRI23c1i2Xnxe

cpp/src/common/tsblock/tsblock.h
```diff
@@ -206,6 +234,15 @@ class RowIterator {
...
     FORCE_INLINE bool end() { return row_id_ >= tsblock_->row_count_; }
 
+    FORCE_INLINE bool has_next() { return row_id_ < tsblock_->row_count_; }
+
+    FORCE_INLINE uint32_t get_column_count() { return column_count_; }
+
+    FORCE_INLINE TSDataType get_data_type(uint32_t slot_index) {
+        ASSERT(slot_index < column_count_);
+        return tsblock_->vectors_[slot_index]->get_vector_type();
+    }
```
jt2594838: slot_index -> column_index

cpp/src/common/tsfile_common.h
```diff
@@ -907,233 +910,251 @@ static const char *meta_index_node_type_names[5] = {
...
+    std::unordered_map<std::string, std::string> tsfile_properties_;
+    typedef std::unordered_map<std::string, std::shared_ptr<TableSchema>>
+        TableSchemasMap;
+    TableSchemasMap table_schemas_;
+    int64_t meta_offset_;
+    BloomFilter *bloom_filter_;
+    common::PageArena *page_arena_;
+
+    int get_table_metaindex_node(common::String &table_name,
+                                 MetaIndexNode *&ret_node) {  // TODO
```
jt2594838: TODO what

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -0,0 +1,304 @@
...
+    if (!last_block_returned_) {
+        return true;
+    }
+
+    if (field_column_contexts_.empty()) {
+        return false;
+    }
+    current_block_->reset();
+
+    next_time_ = -1;
```
jt2594838, line 107: Notice that timestamps can be minus and Int64.MIN_VALUE is also a valid timestamp.
Therefore, you'd better use a flag to indicate if the first timestamp is set.

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -0,0 +1,304 @@
...
+            delete value_iter_;
+            value_iter_ = nullptr;
+        }
+        if (tsblock_) {
+            ssi_->destroy();
+            tsblock_ = nullptr;
+        }
+    } else {
+        std::cout << "debug: \n";
+        std::cout << tsblock_->debug_string() << std::endl;
```
jt2594838: Add debug macro or remove

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -0,0 +1,304 @@
...
+}
+
+int SingleMeasurementColumnContext::get_current_value(char* value) {
+    if (value_iter_->end()) {
+        return common::E_NO_MORE_DATA;
+    }
+    uint32_t len = 0;
+    value = value_iter_->read(&len);
+    return common::E_OK;
+}
```
jt2594838: How can the caller know the size of value?
And should value be `char* &value`?

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -0,0 +1,304 @@
...
+    if (!get_current_value(val)) {
+        return;
+    }
+    for (int32_t pos : pos_in_result_) {
+        int len = 0;
+        if (get_current_value(val)) {
+            col_appenders[pos]->append(val, len);
+        }
+    }
+}
```
jt2594838: len is not set?

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -0,0 +1,304 @@
...
+void SingleDeviceTsBlockReader::construct_column_context(
+    const ITimeseriesIndex* time_series_index, Filter* time_filter) {
+    // TODO: judge whether the time_series_index is aligned and jump empty chunk
+    SingleMeasurementColumnContext* column_context =
+        new SingleMeasurementColumnContext(tsfile_io_reader_);
+    column_context->init(device_query_task_, time_series_index, time_filter,
+                         pa_);
+    field_column_contexts_.insert(std::make_pair(
+        time_series_index->get_measurement_name().to_std_string(),
+        column_context));
```
jt2594838: This todo is necessary

cpp/src/reader/table_result_set.cc
```diff
@@ -0,0 +1,102 @@
...
+    }
+    if (row_iterator_ == nullptr || !row_iterator_->has_next()) {
+        return false;
+    }
+    row_iterator_->next();
+    uint32_t len = 0;
+    bool null = false;
+    for (uint32_t i = 0; i < row_iterator_->get_column_count(); ++i) {
+        std::cout << "TableResultSet::next" << std::endl;
+        std::cout << "data_types_[i]:" << static_cast<int>(row_iterator_->get_data_type(i)) << std::endl;
```
jt2594838: Add debug macro

jt2594838 (changes requested):

cpp/src/common/allocator/my_string.h
```diff
@@ -35,6 +35,13 @@ struct String {
 
     String() : buf_(nullptr), len_(0) {}
     String(char *buf, uint32_t len) : buf_(buf), len_(len) {}
+    String(const std::string& str, common::PageArena& pa) : buf_(nullptr), len_(0) {
+        dup_from(str, pa);
+    }
+    String(const std::string& str) {
+        buf_ = (char*)str.c_str();
+        len_ = str.size();
+    }
```
ColinLeeo, line 44: What this func used for?

cpp/src/common/schema.h
```diff
@@ -20,319 +20,342 @@
...
+                if (to_lower(column_schemas_[i]->measurement_name_) ==
+                    lower_case_column_name) {
+                    index = static_cast<int>(i);
+                    break;
                 }
-                column_pos_index_[lower_case_column_name] = index;
-                return index;
             }
+            column_pos_index_[lower_case_column_name] = index;
+            return index;
```
ColinLeeo: Need to check all places where column_pos_index_ is used, as it may lead to unoccupied space for non-existent column_name not being released.

cpp/src/common/schema.h
ColinLeeo: find_colum_index may return -1 which will lead to vector overflow.

cpp/src/common/schema.h
```diff
@@ -20,319 +20,342 @@
...
+enum class ColumnCategory { TAG = 0, FIELD = 1 };
+
+class TableSchema {
+   public:
+    static void to_lowercase_inplace(std::string &str) {
+        std::transform(
+            str.begin(), str.end(), str.begin(),
+            [](unsigned char c) -> unsigned char { return std::tolower(c); });
+    }
+
```
ColinLeeo: move to private like to_lower or move to utils

cpp/src/common/tsblock/tsblock.h
```diff
@@ -95,6 +103,16 @@ class TsBlock {
...
     }
 
+    FORCE_INLINE static TsBlock *create_tsblock(TupleDesc *tupledesc,
+                                                uint32_t max_row_count = 0) {
+        TsBlock *tsblock = new TsBlock(tupledesc, max_row_count);
+        if (IS_FAIL(tsblock->init())) {
+            delete tsblock;
+            tsblock = nullptr;
+        }
+        return tsblock;
```
ColinLeeo: This func will return nullptr， which will case other modules crash.
Check usage in other modules or return error code.

cpp/src/common/tsblock/tsblock.h
```diff
@@ -178,6 +196,16 @@ class ColAppender {
...
+    FORCE_INLINE bool fill(const char *value, uint32_t len,
+                           uint32_t end_index) {
+        while (column_row_count_ < end_index) {
+            if (!add_row()) {
+                return false;
+            }
+            append(value, len);
+        }
+        return true;
+    }
```
ColinLeeo: The return is not used. Use void instead.

cpp/src/common/tsfile_common.h
```diff
@@ -58,847 +58,850 @@ typedef int64_t TsFileID;
...
+        }
+    }
+    IMetaIndexEntry() = default;
+    virtual ~IMetaIndexEntry() = default;
+
+    virtual int serialize_to(common::ByteStream &out) { return common::E_OK; }
+    virtual int deserialize_from(common::ByteStream &out,
+                                 common::PageArena *pa) {
+        return common::E_OK;
+    }
```
ColinLeeo: For these virtual functions return not_support will be better.

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -0,0 +1,304 @@
...
+    col_appenders_.resize(tuple_desc_.get_column_count());
+    for (int i = 0; i < tuple_desc_.get_column_count(); i++) {
+        col_appenders_[i] = new common::ColAppender(i, current_block_);
+    }
+    row_appender_ = new common::RowAppender(current_block_);
+    std::vector<ITimeseriesIndex*> time_series_indexs(
+        device_query_task_->get_column_names().size());
+    tsfile_io_reader_->get_timeseries_indexes(
+        device_query_task->get_device_id(),
+        device_query_task->get_column_names(), time_series_indexs, pa_);
```
ColinLeeo: check the return code. time_series_indexs may contain nullptr.

ColinLeeo (changes requested):

cpp/src/reader/block/device_ordered_tsblock_reader.cc
```diff
@@ -0,0 +1,76 @@
...
+    }
+    if (time_filter_ != nullptr) {
+        delete time_filter_;
+        time_filter_ = nullptr;
+    }
+    if (field_filter_ != nullptr) {
+        delete field_filter_;
+        field_filter_ = nullptr;
+    }
+}
```
ColinLeeo, line 80: Filter *time_filter_ = nullptr;
    Filter *field_filter_ = nullptr;
    
    these pointers are passed by user, maybe better to free outside.

cpp/src/reader/block/device_ordered_tsblock_reader.cc
```diff
@@ -0,0 +1,76 @@
...
+        if (current_reader_) {
+            delete current_reader_;
+            current_reader_ = nullptr;
+        }
+        current_reader_ = new SingleDeviceTsBlockReader(
+            task, block_size_, metadata_querier_, tsfile_io_reader_, time_filter_,
+            field_filter_);
+        if (current_reader_->has_next()) {
+            return true;
+        }
```
ColinLeeo: How about oom?

cpp/src/common/allocator/my_string.h
zwhzzz0821: SingleDeviceTsBlockReader::fill_ids().
I think we need a temporary String object to write to. So I added this constructor.

cpp/src/reader/block/device_ordered_tsblock_reader.cc
zwhzzz0821: `    int query(const std::string &table_name,
              const std::vector<std::string> &columns_names, int64_t start_time,
              int64_t end_time, ResultSet *&result_set);`
In this interface, we construct a time_filter internally after passing in the time, so I think it should be destructed internally as well.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -384,6 +386,99 @@ int TsFileWriter::do_check_schema_aligned(
...
+        schemas_[device_id] = device_schema;
+    }
+
+    uint32_t column_cnt = tablet.get_column_count();
+    if (tablet.column_categories_.empty()) {
+        for (uint32_t i = 0; i < column_cnt; i++) {
+            auto& col_name = tablet.get_column_name(i);
+            int col_index = table_schema->find_column_index(col_name);
+            const common::ColumnCategory column_category = table_schema->get_column_categories()[col_index];
+            tablet.column_categories_.emplace_back(column_category);
```
ColinLeeo: If col_name is not exist in table_schema, col_index will be -1, which will lead to overflow.

cpp/src/reader/block/device_ordered_tsblock_reader.cc
ColinLeeo: It's ok, when has_next is false, returning common::NO_MORE_DATA would be a better approach, just do it in future.

cpp/src/writer/tsfile_writer.cc
jt2594838: Fixed

ColinLeeo (approved):

jt2594838 (approved):

PR #412, opened 2025-02-18
Fix float encoder overflow when float value itself over int range

HTHou:


java/tsfile/src/main/java/org/apache/tsfile/encoding/decoder/FloatDecoder.java
```diff
@@ -97,6 +99,10 @@ public FloatDecoder(TSEncoding encodingType, TSDataType dataType) {
   public float readFloat(ByteBuffer buffer) {
     readMaxPointValue(buffer);
     int value = decoder.readInt(buffer);
+    if (valueItselfOverflowInfo != null && valueItselfOverflowInfo.isMarked(position)) {
+      position++;
+      return BytesUtils.bytesToFloat(BytesUtils.intToBytes(value));
```
jt2594838: May use the native methods Float.intBitsToFloat() and Float.floatToIntBits(), which are possibly more efficient.

jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/encoding/decoder/FloatDecoder.java
HTHou: Fixed

java/tsfile/src/main/java/org/apache/tsfile/encoding/encoder/FloatEncoder.java
```diff
@@ -137,8 +136,13 @@ private int convertFloatToInt(float value) {
 
   private long convertDoubleToLong(double value) {
     if (value * maxPointValue > Long.MAX_VALUE || value * maxPointValue < Long.MIN_VALUE) {
-      underflowFlags.add(false);
-      return Math.round(value);
+      if (value > Long.MAX_VALUE || value < Long.MIN_VALUE) {
+        underflowFlags.add(null);
+        return Double.doubleToLongBits(value);
```
jt2594838, line 146: Should also consider NaN.

jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/encoding/encoder/FloatEncoder.java
HTHou: Fixed

jt2594838 (approved):

PR #415, opened 2025-02-25
compaction adapting new type when table alter column type

zhujt20:
compaction adapting new type when table alter column type


java/tsfile/src/main/java/org/apache/tsfile/file/metadata/AbstractAlignedChunkMetadata.java
```diff
@@ -160,6 +160,17 @@ public TSDataType getDataType() {
     return timeChunkMetadata.getDataType();
   }
 
+  @Override
+  public TSDataType getNewType() {
+    return timeChunkMetadata.getNewType();
+  }
```
jt2594838: Also throw exception here.

java/tsfile/src/main/java/org/apache/tsfile/read/common/Chunk.java
```diff
@@ -200,4 +214,164 @@ public Statistics getChunkStatistic() {
...
   }
+
+  public Chunk rewrite(TSDataType newType, boolean isValueChunk, Chunk timeChunk)
+      throws IOException {
+    if (newType == null || newType == chunkHeader.getDataType()) {
+      return this;
+    }
+    IMeasurementSchema schema =
+        new MeasurementSchema(
+            chunkHeader.getMeasurementID(), newType, chunkHeader.getEncodingType());
```
jt2594838: Do not forget compression.

java/tsfile/src/main/java/org/apache/tsfile/read/common/Chunk.java
```diff
@@ -200,4 +214,164 @@ public Statistics getChunkStatistic() {
...
+          new ValueChunkWriter(
+              chunkHeader.getMeasurementID(),
+              chunkHeader.getCompressionType(),
+              newType,
+              chunkHeader.getEncodingType(),
+              schema.getValueEncoder(),
+              encryptParam);
+      List<Chunk> valueChunks = new ArrayList<>();
+      valueChunks.add(this);
+      AlignedChunkReader chunkReader = new AlignedChunkReader(timeChunk, valueChunks);
```
jt2594838:
```suggestion
      AlignedChunkReader chunkReader = new AlignedChunkReader(timeChunk, Collections.singletonList(this));
```

jt2594838: Also, since AlignedChunkReader will ignore timestamps that all values are null, please use TableChunkReader, which will not. 
Otherwise, the new value chunk will be unable to align with the time chunk because of missing timestamps.

java/tsfile/src/main/java/org/apache/tsfile/read/common/Chunk.java
```diff
@@ -200,4 +214,164 @@ public Statistics getChunkStatistic() {
...
+      ByteBuffer newChunkData = chunkWriter.getByteBuffer();
+      ChunkHeader newChunkHeader =
+          new ChunkHeader(
+              chunkHeader.getMeasurementID(),
+              newChunkData.capacity(),
+              newType,
+              chunkHeader.getCompressionType(),
+              chunkHeader.getEncodingType(),
+              chunkWriter.getNumOfPages(),
+              TsFileConstant.VALUE_COLUMN_MASK);
```
jt2594838: chunkType should inherit from this one.

java/tsfile/src/main/java/org/apache/tsfile/read/common/Chunk.java
```diff
@@ -200,4 +214,164 @@ public Statistics getChunkStatistic() {
...
+
+  public Chunk rewrite(TSDataType newType, boolean isValueChunk, Chunk timeChunk)
+      throws IOException {
+    if (newType == null || newType == chunkHeader.getDataType()) {
+      return this;
+    }
+    IMeasurementSchema schema =
+        new MeasurementSchema(
+            chunkHeader.getMeasurementID(), newType, chunkHeader.getEncodingType());
+    if (isValueChunk) {
```
jt2594838: Better to split into two methods.

java/tsfile/src/main/java/org/apache/tsfile/read/common/Chunk.java
```diff
@@ -200,4 +214,164 @@ public Statistics getChunkStatistic() {
...
+      BatchData batchData = page.getAllSatisfiedPageData();
+      IPointReader pointReader = batchData.getBatchDataIterator();
+      while (pointReader.hasNextTimeValuePair()) {
+        TimeValuePair point = pointReader.nextTimeValuePair();
+        Object convertedValue =
+            newType.castFromSingleValue(chunkHeader.getDataType(), point.getValue().getValue());
+        long timestamp = point.getTimestamp();
+        if (convertedValue == null) {
+          continue;
+        }
```
jt2594838: In what case convertedValue will be a null？

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/ChunkWriterImpl.java
```diff
@@ -527,4 +527,21 @@ public void setLastPoint(boolean isLastPoint) {
...
+
+  public int getNumOfPages() {
+    return numOfPages;
+  }
+
+  public ByteBuffer getByteBuffer() {
+    ByteBuffer copy = ByteBuffer.allocate(pageBuffer.size());
+    copy.put(pageBuffer.toByteArray());
+    copy.flip();
+    return copy;
```
jt2594838:
```suggestion
    return ByteBuffer.wrap(pageBuffer.toByteArray());
```

jt2594838: The same for other places.

java/tsfile/src/test/java/org/apache/tsfile/write/ChunkRewriteTest.java
```diff
@@ -0,0 +1,189 @@
...
+      int i = 1;
+      while (pointReader.hasNextTimeValuePair()) {
+        TimeValuePair point = pointReader.nextTimeValuePair();
+        assertEquals((long) i, point.getTimestamp());
+        assertEquals((float) i, point.getValue().getVector()[0].getValue());
+        assertEquals((double) i, point.getValue().getVector()[1].getValue());
+        assertEquals((double) i, point.getValue().getVector()[2].getValue());
+        i++;
+      }
+    }
```
jt2594838: Check the number of points.

java/tsfile/src/test/java/org/apache/tsfile/write/ChunkRewriteTest.java
```diff
@@ -0,0 +1,189 @@
...
+      int i = 1;
+      while (pointReader.hasNextTimeValuePair()) {
+        TimeValuePair point = pointReader.nextTimeValuePair();
+        assertEquals((long) i, point.getTimestamp());
+        assertEquals((double) i, point.getValue().getValue());
+        i++;
+      }
+    }
+  }
+}
```
jt2594838, line 425: Add three tests:
1. aligned with more than one page;
2. non-aligned with more than one page;
3.  aligned with some nulls.

jt2594838 (approved):

java/tsfile/src/main/java/org/apache/tsfile/read/common/Chunk.java
zhujt20: Actually, I'm not sure about how nonAlignedChunk deal with null value. Will I read something like (timestamp, null) so when I convert the value there's a null-pointer thrreat?

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/ChunkWriterImpl.java
zhujt20: Actually I write this mainly for get rid of the relation between the new chunk and the new chunkwriter so that the chunkwriter can be destructed quickly. If it's not important, I'll change it.

jt2594838: This does not matter. `pageBuffer.toByteArray()` already copies the data.

java/tsfile/src/main/java/org/apache/tsfile/read/common/Chunk.java
jt2594838: NonAlignedChunk should not contain any null. You may add an error log here, just in case.

jt2594838 (approved):

PR #428, opened 2025-03-04
Fix bugs and optimize TS2diff encoding.

hongzhi-gao:
Fix the inconsistency issues between the files of different Java versions.
Fix the issue of reading large pages.
Enable ts2diff to support SIMD (Single Instruction, Multiple Data).


cpp/src/compress/snappy_compressor.cc
```diff
@@ -26,12 +26,10 @@ namespace storage {
...
     if (compressed_buf_ != nullptr) {
-        mem_free(compressed_buf_);
         compressed_buf_ = nullptr;
     }
 
     if (uncompressed_buf_ != nullptr) {
-        mem_free(uncompressed_buf_);
         uncompressed_buf_ = nullptr;
     }
 }
```
jt2594838: Why are these removed?

jt2594838 (approved):

jt2594838 (approved):

cpp/src/compress/snappy_compressor.cc
761417898: Removed

hongzhi-gao: The changes have been removed.

PR #433, opened 2025-03-08
Fix py read error

ColinLeeo:
In this pr:
1. fix data losing when query a table that not all devices have data.
2. add support for py read table schema.
3. add support for py read dataframe.
4. fix some issues.

### DIFF
--- a/cpp/src/reader/block/device_ordered_tsblock_reader.cc
+++ b/cpp/src/reader/block/device_ordered_tsblock_reader.cc
@@ -32,8 +32,9 @@ int DeviceOrderedTsBlockReader::has_next(bool &has_next) {
     }
     while (device_task_iterator_->has_next()) {
         DeviceQueryTask *task = nullptr;
+        // There is no device anymore, just return has_next = false.
         if (IS_FAIL(device_task_iterator_->next(task))) {
-            has_next = false;
+            return ret;
         }
         if (current_reader_) {
             delete current_reader_;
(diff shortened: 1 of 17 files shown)


HTHou (approved):
Checked the python part. LGTM.

cpp/src/reader/block/device_ordered_tsblock_reader.cc
```diff
@@ -43,15 +43,23 @@ int DeviceOrderedTsBlockReader::has_next(bool &has_next) {
...
+
         if (RET_FAIL(current_reader_->has_next(has_next))) {
             return ret;
-        } else {
-            return common::E_OK;
+        }
+        // If current device do not have data, get next device.
+        if (has_next) {
+            return ret;
+        }
```
jt2594838: The comment and the condition conflict, double-check.

python/tsfile/tsfile_py_cpp.pyx
```diff
@@ -370,6 +387,8 @@ cdef ResultSet tsfile_reader_query_table_c(TsFileReader reader, object table_nam
             columns[i] = strdup((<str>column_list[i]).encode('utf-8'))
             if columns[i] == NULL:
                 raise MemoryError("Failed to allocate memory for column name")
+        print("begin to query table:" + table_name + "start_time:" + str(start_time)
+              +"end_time:" + str(end_time))
```
jt2594838: Remove

jt2594838 (approved):

jt2594838 (approved):

PR #434, opened 2025-03-10
Refactor header packaging & fix zlib missing header

jt2594838:
Reduce the number or copying header targets.
Fix that Zlib will auto rename zconf.h, which causes a missing header.


cpp/src/parser/CMakeLists.txt
```diff
@@ -22,4 +22,5 @@ file(GLOB_RECURSE PARSER_SRC_LIST "*.cpp")
 add_library(parser_obj OBJECT ${PARSER_SRC_LIST})
 
 file(GLOB_RECURSE HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
-copy_to_dir(${HEADERS})
\ No newline at end of file
+SET(PARENT "parser_obj")
```
ColinLeeo: Useless code ?

cpp/third_party/zlib-1.2.13/CMakeLists.txt
```diff
@@ -67,7 +67,8 @@ endif()
 if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
     # If we're doing an out of source build and the user has a zconf.h
     # in their source tree...
-    if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/zconf.h)
+    # only rename in macOS because it conflicts with the embedded header
+    if(APPLE AND EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/zconf.h)
```
ColinLeeo: May be is unnecessary.

cpp/third_party/zlib-1.2.13/zconf.h
ColinLeeo, line 1: This file is generated by zlib, might be unnecessary.

jt2594838: Well, actually it is not. We just removed it some time ago.

cpp/third_party/zlib-1.2.13/CMakeLists.txt
jt2594838: I may try to revert it, but if the CI issue reappears, I will add it back.

cpp/src/parser/CMakeLists.txt
jt2594838: Removed.

PR #446, opened 2025-03-20
fix writer add memory threshold.

ColinLeeo:


jt2594838 (approved):
Better to add a test to check whether the config takes effect.

PR #447, opened 2025-03-23
add getter for encryptParam

zhujt20:
add getter for encryptParam
only initialize encryptUtils.encryptParam when first used


java/tsfile/src/main/java/org/apache/tsfile/encrypt/EncryptUtils.java
```diff
@@ -106,78 +122,86 @@ public static String byteArrayToHexString(byte[] bytes) {
...
+
+    StringBuilder valueStr = new StringBuilder();
+
+    for (byte b : data_key) {
+      valueStr.append(b).append(",");
+    }
+
+    valueStr.deleteCharAt(valueStr.length() - 1);
+    return valueStr.toString();
   }
```
jt2594838: What is the difference between this method and `getNormalKeyStr(TSFileDescriptor.getInstance().getConfig())`?

jt2594838 (approved):

PR #470, opened 2025-04-23
Optimizations regarding chunk metadata sort & timeseries metadata serialization

jt2594838:
When closing a TsFile:
1. The sort of chunk metadata is refactored to reduce redundant calculation;
2. Introduce no-sync streams to reduce synchronization cost during serializing TimeSeriesMetadata;
3. Remove zeroization in  [DeltaBinaryEncoder].

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/TimeseriesMetadata.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/TimeseriesMetadata.java
@@ -21,6 +21,8 @@ package org.apache.tsfile.file.metadata;
 
 import org.apache.tsfile.common.conf.TSFileConfig;
 import org.apache.tsfile.enums.TSDataType;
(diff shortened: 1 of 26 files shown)


java/tsfile/src/main/java/org/apache/tsfile/file/metadata/TimeseriesMetadata.java
```diff
@@ -394,4 +400,37 @@ public String toString() {
         + chunkMetadataList
         + '}';
   }
+
+  public static void main(String[] args) throws IOException {
```
shuwenwei, line 404: Should this test be deleted?

jt2594838: Well, the serialization still accounts for a significant part in my experiment, so I would like to keep this a little longer in case I find some further optimizations in the future.

shuwenwei (approved):

PR #472, opened 2025-04-23
Colin fix config

ColinLeeo:
In this pr:
1. we enable users to specify the memory threshold for flushing when creating writers, 
2. allow users to view and modify internal configurations.


jt2594838 (commented):
Well, I think we need more configs than these, may refer to https://timechor.feishu.cn/docx/P8E4dERvpoMrKfx8jSbcUndxnNc?from=from_copylink

PR #487, opened 2025-05-07
Add flush method for tsfile python writer

ycycse:
We found the automatically flush in write_table() method in tsfileTableWriter doesn't work sometimes which leading to huge tsfile with only one chunk.
This PR introduce flush() method to tsfileTableWriter to avoid such situation by calling this method by hand. 
The implementation shows the code after this PR provides a way to split chunk:
```python
        for i in range(length):
            tablet.add_timestamp(index, i)
            tablet.add_value_by_index(0, index, i * 1.0)
            index += 1
            if index == 100000:
                print("write tablet")
                tsfile_writer.write_table(tablet)
                tsfile_writer.flush()
                tablet = Tablet(column_name, column_datat_type, 100000)
                index = 0
```
Before(without flush):
image
After(use flush):
image


cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -163,6 +163,18 @@ ERRNO tsfile_writer_close(TsFileWriter writer) {
     return ret;
 }
 
+ERRNO tsfile_writer_flush(TsFileWriter writer) {
```
ColinLeeo: This function is not a api for C, remove.

cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -683,6 +695,16 @@ ERRNO _tsfile_writer_close(TsFileWriter writer) {
     return ret;
 }
 
+ERRNO _tsfile_writer_flush(TsFileWriter writer) {
+    auto *w = static_cast<storage::TsFileWriter *>(writer);
+    int ret = w->flush();
+    if (ret != common::E_OK) {
+        return ret;
+    }
```
ColinLeeo: just return

python/tsfile/tsfile_table_writer.py
```diff
@@ -60,6 +60,13 @@ def close(self):
...
         self.writer.close()
 
+    def flush(self):
+        """
+        Flush current data to tsfile.
+        :return: no return value.
+        """
+        self.writer.flush()
+
     def __dealloc__(self):
```
ColinLeeo, line 70: Add some tests.

PR #489, opened 2025-05-12
Fix tag empty error and disorder timestamp.

ColinLeeo:
In this pr:
Fixed the issue of out-of-order data writes.
Fixed the issue related to writing and reading with empty tags.
Added support for writing null tag values.


cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -199,9 +199,15 @@ int SingleDeviceTsBlockReader::fill_ids() {
...
+                id_column_context.pos_in_device_id_) {
+                device_id = common::String("");
+            } else {
+                device_id = common::String(
+                    device_query_task_->get_device_id()->get_segments().at(
+                        id_column_context.pos_in_device_id_));
+            }
             if (RET_FAIL(col_appenders_[pos + 1]->fill(
                     (char*)&device_id, sizeof(device_id),
                     current_block_->get_row_count()))) {
```
jt2594838: We should distinguish null from empty string.
("table1", null, "tag1") is different from ("table1", "", "tag1")

jt2594838 (approved):

cpp/src/common/allocator/byte_stream.h
```diff
@@ -1047,6 +1048,50 @@ class SerializationUtil {
...
     }
+
+    // If the str is nullptr, NO_STR_TO_READ will be added instead.
+    FORCE_INLINE static int write_var_char_ptr(const char* str, ByteStream &out) {
+        int ret = common::E_OK;
+        if (str == nullptr) {
+            write_var_int(storage::NO_STR_TO_READ, out);
+            return ret;
+        }
+        size_t str_len = std::strlen(str);
```
jt2594838: It is not guaranteed that a tag will not contain '\0', so it is invalid to use std::strlen.

cpp/src/common/device_id.h
```diff
@@ -35,25 +35,25 @@
...
 class IDeviceID {
-public:
+   public:
     virtual ~IDeviceID() = default;
     virtual int serialize(common::ByteStream& write_stream) { return 0; }
     virtual int deserialize(common::ByteStream& read_stream) { return 0; }
     virtual std::string get_table_name() { return ""; }
     virtual int segment_num() { return 0; }
-    virtual const std::vector<std::string>& get_segments() const {
+    virtual const std::vector<char*>& get_segments() const {
```
jt2594838: As stated, C-style strings may have a potential problem.
How about using std::string*?

cpp/src/common/allocator/byte_stream.h
```diff
@@ -1078,14 +1080,17 @@ class SerializationUtil {
                 str = nullptr;
                 return ret;
             } else {
-                char *tmp_buf = (char*) malloc(len + 1);
+                char *tmp_buf = static_cast<char *>(malloc(len + 1));
                 tmp_buf[len] = '\0';
```
jt2594838: Is adding '\0' still necessary?

jt2594838 (approved):

jt2594838 (approved):

PR #498, opened 2025-05-20
Add TsFileLastReader for retrieving last points in a TsFile

jt2594838:
As the title declares

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/ChunkMetadata.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/ChunkMetadata.java
@@ -430,4 +430,8 @@ public class ChunkMetadata implements IChunkMetadata {
   public MeasurementSchema toMeasurementSchema() {
     return new MeasurementSchema(measurementUid, tsDataType, encoding, compressionType);
   }
+
+  public TSEncoding getEncoding() {
+    return encoding;
+  }
 }
--- a/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/ChunkMetadata.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/ChunkMetadata.java
@@ -430,8 +430,4 @@ public class ChunkMetadata implements IChunkMetadata {
   public MeasurementSchema toMeasurementSchema() {
     return new MeasurementSchema(measurementUid, tsDataType, encoding, compressionType);
   }
-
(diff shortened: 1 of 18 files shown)


shuwenwei (approved):

java/tsfile/src/main/java/org/apache/tsfile/file/metadata/ChunkMetadata.java
```diff
@@ -430,4 +430,8 @@ public boolean hasNullValue(int measurementIndex) {
   public MeasurementSchema toMeasurementSchema() {
     return new MeasurementSchema(measurementUid, tsDataType, encoding, compressionType);
   }
+
+  public TSEncoding getEncoding() {
```
shuwenwei: This method is not used and it may lead to the wrong way of using it

java/tsfile/src/main/java/org/apache/tsfile/read/reader/TsFileLastReader.java
```diff
@@ -115,26 +139,90 @@ public Pair<IDeviceID, List<Pair<String, TimeValuePair>>> next() {
...
+      return new Pair<>(
+          seriesMeta.getMeasurementId(),
+              new TimeValuePair(
+                  seriesMeta.getStatistics().getEndTime(),
+                  seriesMeta.getTsDataType() == TSDataType.VECTOR ?
+                      TsPrimitiveType.getByType(TSDataType.INT64, seriesMeta.getStatistics().getEndTime()) :
+                  TsPrimitiveType.getByType(seriesMeta.getTsDataType(),
+                      seriesMeta.getStatistics().getLastValue())));
+    } else {
+      ChunkMetadata chunkMetadata = (ChunkMetadata) seriesMeta.getChunkMetadataList()
```
shuwenwei: The last chunk could be an empty chunk

java/tsfile/src/main/java/org/apache/tsfile/file/metadata/ChunkMetadata.java
jt2594838: Removed

java/tsfile/src/main/java/org/apache/tsfile/read/reader/TsFileLastReader.java
jt2594838: Fixed

shuwenwei (approved):

PR #514, opened 2025-06-16
[To dev/1.1] Correct the retained size calculation for BinaryColumn and BinaryColumnBuilder

JackieTien97:
cp from https://github.com/apache/tsfile/pull/384


.mvn/.develocity/develocity-workspace-id
HTHou, line 1: Can be removed?

JackieTien97: fixed

PR #550, opened 2025-07-21
Add a method to batch search for the MetadataIndexNode offset of devices

shuwenwei:
For a tsfile with 80,000 devices, the time to read metadata can be reduced from 4.5s to 500ms.

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/MetadataIndexNode.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/file/metadata/MetadataIndexNode.java
@@ -264,8 +264,7 @@ public class MetadataIndexNode {
         i++;
       }
     }
-    Arrays.fill(
-        results, i, keys.size(), exactSearch ? -1 : children.size() - 1);
+    Arrays.fill(results, i, keys.size(), exactSearch ? -1 : children.size() - 1);
     return results;
   }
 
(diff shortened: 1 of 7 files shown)


java/tsfile/src/main/java/org/apache/tsfile/read/TsFileSequenceReader.java
```diff
@@ -535,6 +535,82 @@ private Map<String, TimeseriesMetadata> readDeviceMetadataFromDisk(IDeviceID dev
...
+   * internal MetadataIndexNode
+   *
+   * @param table table name, or "" for tree model
+   * @param sortedDevices devices should be sorted
+   * @param ioSizeRecorder can be null
+   * @return Each element of the outer array corresponds to the device at this index. The inner
+   *     array size is 2, the first element is the start offset, and the second is the end offset
+   * @throws IOException io error
+   */
+  public long[][] getDeviceMetadataIndexNodeOffsets(
```
JackieTien97: using long[] may be better for memory consumption?

java/tsfile/src/main/java/org/apache/tsfile/file/metadata/MetadataIndexNode.java
```diff
@@ -33,10 +33,12 @@
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 public class MetadataIndexNode {
 
+  private static final double LOG2 = Math.log(2);
```
JackieTien97, line 41: what's this magic number used for?

JackieTien97 (changes requested):

java/tsfile/src/main/java/org/apache/tsfile/file/metadata/MetadataIndexNode.java
shuwenwei: Used for calculating complexity, it will select the search method according to the complexity.

PR #613, opened 2025-10-29
Object type

HTHou:


java/tsfile/src/main/java/org/apache/tsfile/read/common/type/ObjectType.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public boolean isComparable() {
+    return true;
+  }
+
+  @Override
+  public boolean isOrderable() {
+    return true;
```
JackieTien97:
```suggestion
    return false;
```
we think object type is not orderable, that means it can't used in order by clause

java/common/src/main/java/org/apache/tsfile/enums/TSDataType.java
```diff
@@ -558,14 +579,19 @@ public boolean isComparable() {
...
     }
   }
 
   public boolean isBinary() {
-    return this == TEXT || this == STRING || this == BLOB;
+    return this == TEXT || this == STRING || this == BLOB || this == OBJECT;
+  }
+
+  public boolean isBlob() {
+    return this == BLOB || this == OBJECT;
```
JackieTien97: A little bit confusing? Actually, OBJECT is totaly different from BLOB in their storage. maybe we need to change the function name and add some java doc about what this metod is ued to?

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #636, opened 2025-11-12
Added memory calculation for tablet

Caideyipi:


jt2594838 (approved):

jt2594838 (commented):
remove the two empty files

PR #642, opened 2025-11-19
Fix/cpp get device schema

hongzhi-gao:
Fix the getDeviceSchema interface in TsFile-CPP to resolve the issue of only returning the first MeasurementSchema.


ColinLeeo (approved):

ColinLeeo:
Good Job

PR #648, opened 2025-11-27
Feature/cpp removed antlr4

hongzhi-gao:
**Library size reduced from 3.2 MB to 1.7 MB by making dependencies optional.**
1. Support building tsfile-cpp without depending on ANTLR4/snappy/lz4/lzo/gzip, reducing package size
2. Support building without dependencies on ANTLR4 and GTest, avoiding dependency download failures due to network issues

Before
img_v3_02se_981ed99c-0073-4093-9459-23ab2d63fdbg

**Package size after removing ANTLR4**
image

**Package size after removing ANTLR4 and compression algorithms (Snappy/LZ4/LZOKAY/Gzip)**
image


cpp/src/common/device_id.cc
```diff
@@ -0,0 +1,396 @@
...
+                           [](const std::string& seg) { return !seg.empty(); });
+    std::vector<std::string> validate_segments(segments.begin(), it.base());
+    std::vector<std::string*> result;
+    result.reserve(validate_segments.size());
+    for (const auto& segment : validate_segments) {
+        result.emplace_back(new std::string(segment));
+    }
+    return result;
+}
+
```
jt2594838, line 207: Null is different from "", so you cannot remove trailing empty segments.

cpp/src/common/device_id.cc
```diff
@@ -0,0 +1,396 @@
...
+std::vector<std::string> IDeviceID::split_string(const std::string& str,
+                                                 char delimiter) {
+    std::vector<std::string> tokens;
+
+    // Reject newlines in path explicitly (illegal path name).
+    if (str.find('\n') != std::string::npos ||
+        str.find('\r') != std::string::npos) {
+        throw std::runtime_error("Path contains newline");
+    }
```
jt2594838, line 270: No need to ban these?

cpp/src/common/device_id.cc
```diff
@@ -0,0 +1,396 @@
...
+}
+
+std::vector<std::string> IDeviceID::split_string(const std::string& str,
+                                                 char delimiter) {
+    std::vector<std::string> tokens;
+
+    // Reject newlines in path explicitly (illegal path name).
+    if (str.find('\n') != std::string::npos ||
+        str.find('\r') != std::string::npos) {
+        throw std::runtime_error("Path contains newline");
```
jt2594838, line 269: Return tokens as a parameter and return an error code？

cpp/src/common/device_id.cc
```diff
@@ -0,0 +1,396 @@
...
+            token += c;  // preserve
+        } else if (c == '\'' && !in_back_quotes && !in_double_quotes) {
+            in_single_quotes = !in_single_quotes;
+            token += c;  // preserve
+        } else if (c == delimiter && !in_back_quotes && !in_double_quotes &&
+                   !in_single_quotes) {
+            // delimiter outside quotes -> split
+            if (!token.empty()) {
+                tokens.push_back(token);
+                token.clear();
```
jt2594838: verify and unquote？

cpp/src/common/device_id.cc
```diff
@@ -0,0 +1,396 @@
...
+            } else {
+                // keep behaviour: do not push empty tokens
+                token.clear();
+            }
```
jt2594838: Empty segments should be preserved?

cpp/src/compress/compressor_factory.h
```diff
@@ -46,19 +58,35 @@ class CompressorFactory {
...
             ALLOC_AND_RETURN_COMPRESSPR(LZOCompressor);
+#else
+            ALLOC_AND_RETURN_COMPRESSPR(UncompressedCompressor);
+#endif
+        } else if (type == common::LZ4) {
+#ifdef ENABLE_LZ4
+            ALLOC_AND_RETURN_COMPRESSPR(LZ4Compressor);
+#else
+            ALLOC_AND_RETURN_COMPRESSPR(UncompressedCompressor);
+#endif
```
jt2594838, line 83: Make sure that the compression type of the caller is consistently changed.
That is, you will not write a chunk with compression type being LZ4 but the chunk is actually not compressed.

cpp/src/reader/meta_data_querier.h
```diff
@@ -20,6 +20,8 @@
 #ifndef READER_META_DATA_QUERIER_H
 #define READER_META_DATA_QUERIER_H
 
+#include <mutex>
```
jt2594838, line 23: What is this for?

cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -24,6 +24,8 @@
 #include <unistd.h>
 #include <writer/tsfile_table_writer.h>
 
+#include <set>
```
jt2594838, line 27: Is this used?

cpp/src/common/device_id.cc
hongzhi-gao: fixed

cpp/src/common/device_id.cc
hongzhi-gao: removed

cpp/src/common/device_id.cc
hongzhi-gao: preserved

cpp/src/common/device_id.cc
hongzhi-gao: fixed

cpp/src/cwrapper/tsfile_cwrapper.cc
hongzhi-gao: yes

cpp/src/reader/meta_data_querier.h
hongzhi-gao: Removed unused var
```
    std::unique_ptr<
        common::Cache<std::string, /*Todo std::pair<IDeviceID, std::string>*/
                      std::vector<std::shared_ptr<ChunkMeta>>, std::mutex>>
        device_chunk_meta_cache_;
```

cpp/src/compress/compressor_factory.h
hongzhi-gao: fixed.  Return nullptr if not exists

cpp/src/common/device_id.cc
hongzhi-gao: Keeping this consistent with ANTLR4 handling
```
std::vector<std::string> StringArrayDeviceID::split_device_id_string(
    const std::string& device_id_string) {
#ifdef ENABLE_ANTLR4
    auto splits = storage::PathNodesGenerator::invokeParser(device_id_string);
    return split_device_id_string(splits);
#else
    return split_string(device_id_string, '.');
#endif
}
```

jt2594838 (approved):

PR #655, opened 2025-12-04
Replace value decoder

shuwenwei:


java/tsfile/src/main/java/org/apache/tsfile/read/TsFileSequenceReader.java
```diff
@@ -1671,6 +1671,7 @@ public List<TimeseriesMetadata> getDeviceTimeseriesMetadata(IDeviceID device) th
 
   private List<TimeseriesMetadata> getDeviceTimeseriesMetadata(
       IDeviceID device, boolean needChunkMetadata) throws IOException {
+    readFileMetadata();
```
JackieTien97, line 1674: why need to add this line?

JackieTien97 (changes requested):

PR #665, opened 2025-12-16
add the NoMatcheFilter for the case that all the rows don't match

alpass163:
Introduced NoMatchFilter  to handle always-false predicates. This optimization enables the engine to prune impossible conditions early, skipping unnecessary data scanning and reducing I/O overhead.


JackieTien97 (changes requested):
Change the enum name to `FALSE_LITERAL`, Filter name to `FalseLiteralFilter`

JackieTien97 (approved):

PR #676, opened 2025-12-25
Increment the getBinary() function in nonBinary Column

zerolbsony:
Allow other value of data type can be converted to value of Binary data type when find type is not consistence in the column of tsblock.
Resolve this problem :
image


java/tsfile/src/main/java/org/apache/tsfile/read/common/block/column/BooleanColumn.java
```diff
@@ -93,11 +95,26 @@ public boolean getBoolean(int position) {
     return values[position + arrayOffset];
   }
 
+  @Override
+  public Binary getBinary(int position) {
+    Binary[] binaries = getBinaries();
+    return binaries.length > 0 ? binaries[position + arrayOffset] : null;
+  }
+
```
jt2594838, line 102: Not necessary to copy the whole array just because accessing a single value.

jt2594838 (changes requested):

PR #706, opened 2026-01-14
Support dataframe to tsfile

ColinLeeo:
```python
def dataframe_to_tsfile(dataframe: pd.DataFrame,
                        file_path: str,
                        table_name: Optional[str] = None,
                        time_column: Optional[str] = None,
                        tag_column: Optional[list[str]] = None,
                        ):
    """
    Write a pandas DataFrame to a TsFile by inferring the table schema from the DataFrame.
    This function automatically infers the table schema based on the DataFrame's column
    names and data types, then writes the data to a TsFile.
    Parameters
    ----------
    dataframe : pd.DataFrame
        The pandas DataFrame to write to TsFile.
        - If a 'time' column (case-insensitive) exists, it will be used as the time column.
        - Otherwise, the DataFrame index will be used as timestamps.
        - All other columns will be treated as data columns.
    file_path : str
        Path to the TsFile to write. Will be created if it doesn't exist.
    table_name : Optional[str], default None
        Name of the table. If None, defaults to "table".
    time_column : Optional[str], default None
        Name of the time column. If None, will look for a column named 'time' (case-insensitive),
        or use the DataFrame index if no 'time' column is found.
    tag_column : Optional[list[str]], default None
        List of column names to be treated as TAG columns. All other columns will be FIELD columns.
        If None, all columns are treated as FIELD columns.
    Returns
    -------
    None
    Raises
    ------
    ValueError
        If the DataFrame is empty or has no data columns.
    """
```


python/tests/test_dataframe.py
```diff
@@ -0,0 +1,251 @@
...
+
+def test_write_dataframe_all_datatypes():
+    table = TableSchema("test_table",
+                        [ColumnSchema("bool_col", TSDataType.BOOLEAN, ColumnCategory.FIELD),
+                         ColumnSchema("int32_col", TSDataType.INT32, ColumnCategory.FIELD),
+                         ColumnSchema("int64_col", TSDataType.INT64, ColumnCategory.FIELD),
+                         ColumnSchema("float_col", TSDataType.FLOAT, ColumnCategory.FIELD),
+                         ColumnSchema("double_col", TSDataType.DOUBLE, ColumnCategory.FIELD),
+                         ColumnSchema("string_col", TSDataType.STRING, ColumnCategory.FIELD),
+                         ColumnSchema("blob_col", TSDataType.BLOB, ColumnCategory.FIELD)])
```
jt2594838: DATE, TIMESTAMP, and TEXT?

jt2594838 (approved):

cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -687,6 +691,9 @@ ERRNO _tsfile_writer_register_table(TsFileWriter writer, TableSchema* schema) {
     measurement_schemas.resize(schema->column_num);
     for (int i = 0; i < schema->column_num; i++) {
         ColumnSchema* cur_schema = schema->column_schemas + i;
+        if (cur_schema->column_category == TIME) {
+            continue;
+        }
```
jt2594838: If you skip the time column, it will be missing when the file is loaded into IoTDB.

python/tests/test_to_tsfile.py
```diff
@@ -0,0 +1,378 @@
...
+        dataframe_to_tsfile(df, tsfile_path, table_name="test_table", time_column="timestamp")
+
+        df_read = to_dataframe(tsfile_path, table_name="test_table")
+        df_read = df_read.sort_values('time').reset_index(drop=True)
+        df_sorted = convert_to_nullable_types(df.sort_values('timestamp').reset_index(drop=True))
+
+        assert df_read.shape == (30, 3)
+        assert df_read["time"].equals(df_sorted["timestamp"])
+        assert df_read["device"].equals(df_sorted["device"])
+        assert df_read["value"].equals(df_sorted["value"])
```
jt2594838, line 141: The name of the time column should be prevserved.

jt2594838: If the tsfile has defined the time column's name, the generated dataframe should have the same name for the time column.

cpp/src/common/tsblock/tuple_desc.h
```diff
@@ -76,6 +76,15 @@ class TupleDesc {
...
     }
 
+    FORCE_INLINE int get_time_column_index() const {
+        for (uint32_t i = 0; i < column_list_.size(); i++) {
+            if (column_list_[i].get_column_category() == ColumnCategory::TIME) {
+                return i;
+            }
+        }
+        return -1;
+    }
```
jt2594838, line 88: Save the index to avoid iteration every time.

jt2594838 (approved):

PR #730, opened 2026-02-25
Fix empty TAG column result in to_dataframe when querying table model.

ColinLeeo:


python/tsfile/utils.py
```diff
@@ -116,10 +116,16 @@ def _gen(is_iterator: bool) -> Iterator[pd.DataFrame]:
 
             is_tree_model = len(table_schema) == 0
             time_column = None
+            column_name_to_query = []
+            no_data_query = True
             if is_tree_model:
                 if _column_names is None:
                     print("columns name is None, return all columns")
+                # When querying tables in the tree, only measurements are allowed currently.
+                no_data_query = False
```
jt2594838: no_data_query -> no_field_query

jt2594838 (approved):

PR #731, opened 2026-02-27
Feature/get-sensor-statistic-and-restorable-writer

hongzhi-gao:
## Summary

This PR adds **TsFile recovery and continued writing** in C++: open an incomplete or corrupted TsFile, truncate bad tail data, and continue appending with the existing tree and table writers. The behavior is aligned with the Java `RestorableTsFileIOWriter` and enables safe recovery after crashes or interrupted writes.

## Motivation

- Support **recoverable TsFile writes**: detect incomplete/corrupted files, truncate to the last valid offset, and resume writing without losing previously committed data.
- Allow **tail metadata generation** after recovery: recovered chunk group meta and chunk statistics are restored so that `close()` can write correct metadata index and file tail.
- Unify **reader metadata API**: introduce `DeviceTimeseriesMetadataMap` and overloads of `get_timeseries_metadata(device_ids)` / `get_timeseries_metadata()` returning a map; add `get_all_devices()` where appropriate. Only existing devices are included when querying by device list.

### RestorableTsFileIOWriter (new)

- **Open**: `open(file_path, truncate_corrupted)` opens with `O_RDWR|O_CREAT` (no `O_TRUNC`) and runs a **self-check**:
  - Empty file → treat as crashed, allow writing from scratch.
  - Valid header + tail magic → **complete file**; close write handle, `can_write() = false`.
  - Otherwise → **recovery path**: scan from header to find the last valid truncation point, optionally truncate, then re-init the base writer and **restore write_stream_** from file content so `cur_file_position()` is correct when generating tail metadata later.
- **Recovery details**:
  - Rebuild **chunk group meta** and **chunk statistics** (multi-page: merge page stats; single-page: decompress and decode to fill stats; aligned value chunks use time batch from the time chunk in the same group).
  - **write_stream_** is refilled by reading the current file content into the stream; `flush_skip_leading_` is set so that `flush_stream_to_file()` skips these leading bytes and only writes new data. No change to the normal write path.
  - Recovered `ChunkGroupMeta` entries are pushed via `push_chunk_group_meta()` and marked with `chunk_group_meta_from_recovery_` so `destroy()` does not free them (they live in the recovery arena).

### TsFileIOWriter (base)

- **Recovery-only hooks** (defaults keep normal behavior unchanged):
  - `flush_skip_leading_`: when > 0, `flush_stream_to_file()` skips that many leading bytes in the stream (already on disk) and writes the rest.
  - `chunk_group_meta_from_recovery_`: when true, `destroy()` skips freeing chunk group meta (owned by recovery).
- Added `push_chunk_group_meta()`, `set_flush_skip_leading()` (protected) and `friend RestorableTsFileIOWriter`.

### WriteFile

- Added `truncate(size)`, `seek_to_end()`, `get_position()`, and `get_fd()` for recovery (truncate, append position, and reading file content to restore `write_stream_`). `close()` is idempotent when already closed.

### TsFileWriter / TsFileTableWriter / TsFileTreeWriter

- **TsFileWriter**: `init(RestorableTsFileIOWriter* rw)` initializes from a recovered writer: takes schema from the file, does not own `io_writer_` (`io_writer_owned_ = false`). Ensures `time_chunk_writer_` is created when appending after recovery.
- **TsFileTableWriter** / **TsFileTreeWriter**: New constructors that take `RestorableTsFileIOWriter*` for appending after recovery (schema/alignment from restored file).

### Reader API refactor

- **DeviceTimeseriesMetadataMap** type alias: `IDeviceID` → `vector<shared_ptr<ITimeseriesIndex>>`.
- **TsFileReader**: `get_timeseries_metadata(device_ids)` returns a map (only existing devices); `get_timeseries_metadata()` returns metadata for all devices; `get_all_devices()` returns the same as `get_all_device_ids()`.
- **TsFileTreeReader**: Same metadata API; `get_all_devices()` returns `vector<shared_ptr<IDeviceID>>` (underlying reader).

## Testing

- **RestorableTsFileIOWriterTest**: open empty file, bad magic, complete file, truncated file, header-only file; recovery then continue write with **TsFileWriter**, **TsFileTreeWriter** (multi-device), **TsFileTableWriter**, and aligned timeseries; read-back with tree/table reader and assert row counts and device counts.
- **WriteFile**: `TruncateFile` test for `truncate(size)`.
- **TsFileReaderTest / TsFileReaderTreeTest**: Updated to use new `get_timeseries_metadata()` / `get_timeseries_metadata(device_ids)` and `DeviceTimeseriesMetadataMap`; assertions unchanged.


cpp/src/file/restorable_tsfile_io_writer.cc
```diff
@@ -0,0 +1,868 @@
...
+    // to fill out_stat. is_time_column: bit 0x80 in chunk_type_ indicates time
+    // column (aligned model).
+    const bool is_time_column =
+        (static_cast<unsigned char>(chdr.chunk_type_) & 0x80) != 0;
+    PageHeader ph;
+    int ret = ph.deserialize_from(bs, false, chdr.data_type_);
+    if (ret != common::E_OK || ph.compressed_size_ == 0 ||
+        bs.remaining_size() < ph.compressed_size_) {
+        return E_OK;
+    }
```
jt2594838, line 270: Why return E_OK?

cpp/src/file/restorable_tsfile_io_writer.cc
```diff
@@ -0,0 +1,868 @@
...
+                         static_cast<common::TSDataType>(chdr.data_type_),
+                         chunk_start, stat, 0,
+                         static_cast<common::TSEncoding>(chdr.encoding_type_),
+                         static_cast<common::CompressionType>(
+                             chdr.compression_type_),
+                         self_check_arena_);
+                cur_cgm->push(cm);
+                if (cur_device_id != nullptr &&
+                    (static_cast<unsigned char>(chdr.chunk_type_) & 0x80) !=
+                        0) {
```
jt2594838: Add a comment for this or use a meaningful macro instead

cpp/src/file/restorable_tsfile_io_writer.cc
```diff
@@ -0,0 +1,868 @@
...
+                    aligned_devices_.insert(cur_device_id->get_table_name());
```
jt2594838, line 772: Explain this

cpp/src/file/restorable_tsfile_io_writer.cc
```diff
@@ -0,0 +1,868 @@
...
+                break;
+            }
+            offset += nr;
+        }
+        if (ret == E_OK) {
+            set_flush_skip_leading(restored_size);
+        } else {
+            return ret;
+        }
+    }
```
jt2594838, line 816: This seems very expensive. Is not there a simpler way?

cpp/src/writer/tsfile_writer.cc
```diff
@@ -113,11 +115,50 @@ int TsFileWriter::init(WriteFile* write_file) {
...
+    io_writer_ = rw;  // RestorableTsFileIOWriter is also a TsFileIOWriter
+    std::shared_ptr<Schema> known = rw->get_known_schema();
+    for (const auto& kv : known->table_schema_map_) {
+        const std::string& table_name = kv.first;
+        const std::shared_ptr<TableSchema>& ts = kv.second;
+        if (!ts || ts->get_measurement_names().empty()) {
+            continue;
+        }
+        auto device_id = std::make_shared<StringArrayDeviceID>(table_name);
+        auto* ms_group = new MeasurementSchemaGroup;
```
jt2594838: Explain this

jt2594838 (approved):

cpp/src/file/restorable_tsfile_io_writer.cc
hongzhi-gao: Replaced magic number 0x80 with named constant kTimeChunkTypeMask and added comments.

cpp/src/file/restorable_tsfile_io_writer.cc
hongzhi-gao: If recovered chunk is an aligned time chunk, mark this device as aligned to keep post-recovery write behavior consistent.

cpp/src/writer/tsfile_writer.cc
hongzhi-gao: We rebuild mutable in-memory schema cache from recovered schema; table_name is used as device key, and ms_group supports lazy writer creation later.

cpp/src/file/restorable_tsfile_io_writer.cc
hongzhi-gao: fixed. I aligned this branch with Java selfCheck behavior: malformed/incomplete single-page chunk data is now treated as corruption (E_TSFILE_CORRUPTED) instead of being silently skipped.

cpp/src/file/restorable_tsfile_io_writer.cc
hongzhi-gao: fixed  it . I removed the replay loop and added a recovery API to restore the logical write_stream position directly from recovered file size, so we avoid pread+write_buf replay while keeping cur_file_position() correct.

cpp/src/file/restorable_tsfile_io_writer.cc
jt2594838: Then, why is the table_name being instead of the device itself

cpp/src/writer/tsfile_writer.cc
jt2594838: Why "table_name is used as device key"

jt2594838 (approved):

PR #734, opened 2026-03-02
Fix/aligned tv page seal

hongzhi-gao:
**Summary**  

Implement aligned-model synchronized page sealing (matching Java semantics) and fix null-handling bugs in both writer and reader paths.

**Changes**  

1. **Aligned seal sync**: When any aligned page hits its capacity threshold (point count or memory), seal all time/value pages together (`tsfile_writer.cc`).  
2. **All-null value pages**: Allow zero-sized value buffers in `ValuePageData::init`; fix `end_encode_chunk` to correctly handle pre-sealed pages and avoid writing redundant all-null pages in the table model (`value_page_writer.cc`, `value_chunk_writer.cc`, `time_chunk_writer.cc`).  
3. **Table-model seal guard**: Use `get_point_numer() > 0` instead of `has_current_page_data()` in the table write loop to prevent excessive sealing of all-null pages, which caused heap corruption (`0xC0000374`) in `TestNullInTable4`.  
4. **Reader null handling**: Use null-aware `read()` in `QDSWithoutTimeGenerator::next()` to avoid reading past the value buffer for null rows.

**Testing**  

- New tests: `AlignedSealSync_PointCountWithNulls`, `AlignedSealSync_TimeMemoryFirst`, `AlignedSealSync_ValueMemoryFirst`.  
- Full `TsFile_Test` passes in both Release and Debug.


cpp/src/writer/tsfile_writer.cc
```diff
@@ -716,15 +764,33 @@ int TsFileWriter::write_tablet_aligned(const Tablet& tablet) {
...
-            continue;
+    for (uint32_t row = 0; row < tablet.get_cur_row_size(); row++) {
+        int32_t time_pages_before = time_chunk_writer->num_of_pages();
+        std::vector<int32_t> value_pages_before(value_chunk_writers.size(), 0);
+        for (uint32_t c = 0; c < value_chunk_writers.size(); c++) {
+            ValueChunkWriter* value_chunk_writer = value_chunk_writers[c];
+            if (!IS_NULL(value_chunk_writer)) {
+                value_pages_before[c] = value_chunk_writer->num_of_pages();
+            }
+        }
```
jt2594838, line 932: Changing from a column-based order to a row-based order may cause a performance downgrade.
May add a configuration to optionally sacrifice the page size limit:
If strict_page_size=true, use the row-based insertion with checking the page size each time.
Otherwise, insert each column wholly without the page size check, and perform it at the very end.

jt2594838: Or, when there is no string/blob/text column, the memory size of each column can be directly computed with its length. Thus, we can divide the incoming column into splits that will cause a page seal, and write them one by one in a column-based order.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -808,26 +874,38 @@ int TsFileWriter::write_table(Tablet& tablet) {
...
+                int32_t time_pages_after = time_chunk_writer->num_of_pages();
+                if (time_pages_after > time_pages_before) {
+                    for (uint32_t k = 0; k < value_chunk_writers.size(); k++) {
+                        if (!IS_NULL(value_chunk_writers[k]) &&
+                            value_chunk_writers[k]->get_point_numer() > 0 &&
+                            RET_FAIL(
+                                value_chunk_writers[k]->seal_current_page())) {
+                            return ret;
+                        }
                     }
```
jt2594838, line 1388: Why is the number of value pages not checked.

cpp/src/writer/tsfile_writer.cc
hongzhi-gao: Implemented the review suggestion for aligned page sealing.

- Added a new config flag `strict_page_size_` (default: true).
- Extended `TimeChunkWriter` and `ValueChunkWriter` with `set_enable_page_seal_if_full(bool)` so callers can disable the per-`write()` auto page-size/point-number sealing check and seal pages manually at chosen boundaries.

Aligned writing behavior:
- strict_page_size=true: keep the original row-based insertion and use `maybe_seal_aligned_pages_together()` so that when either time or any value side becomes full, all aligned pages are sealed together (strict page size semantics).
- strict_page_size=false:
  - If there are no STRING/TEXT/BLOB columns: switch to column-based insertion. Disable auto sealing for time/value and split by `page_writer_max_point_num_`, sealing pages at segment boundaries.
  - If there are STRING/TEXT/BLOB columns: write time first with auto sealing enabled and record the time page seal row boundaries. Then write each value column with auto sealing disabled and seal manually at those recorded boundaries.

This restores a faster column-based write path while preserving page alignment requirements according to the review proposal.

cpp/src/writer/tsfile_writer.cc
hongzhi-gao: We now mirror write_record_aligned / write_tablet_aligned: each row we snapshot time_pages_before and per-column value_pages_before, then call maybe_seal_aligned_pages_together(), which seals all aligned pages when either time or any value side advances num_of_pages().

hongzhi-gao:
**Force-pushed to clean up an incorrect commit in the previous version.**
**No functional changes intended — only removed an unintended commit and rebased onto latest upstream.**


@[hongzhi-gao](https://github.com/hongzhi-gao) hongzhi-gao [force-pushed](https://github.com/apache/tsfile/compare/d64ab9aa9130b6bdbfd22cccb7862eb14c66a3c0..660f0419a328a4728ba9350692fb64fc6e949b60) the fix/aligned_tv_page_seal branch from [d64ab9a](https://github.com/apache/tsfile/commit/d64ab9aa9130b6bdbfd22cccb7862eb14c66a3c0) to [660f041](https://github.com/apache/tsfile/commit/660f0419a328a4728ba9350692fb64fc6e949b60) 
[21 minutes ago](https://github.com/apache/tsfile/pull/734#event-23922003531)

cpp/src/writer/tsfile_writer.cc
```diff
@@ -764,35 +856,217 @@ int TsFileWriter::write_tablet_aligned(const Tablet& tablet) {
...
+    // (this may sacrifice the strict page size limit for performance).
+    time_chunk_writer->set_enable_page_seal_if_full(true);
+    for (uint32_t c = 0; c < value_chunk_writers.size(); c++) {
+        if (!IS_NULL(value_chunk_writers[c])) {
+            value_chunk_writers[c]->set_enable_page_seal_if_full(false);
+        }
+    }
+
+    std::vector<uint32_t> time_page_row_ends;
+    time_page_row_ends.reserve(total_rows / 16 + 1);
```
jt2594838: What is this 16?

jt2594838 (approved):

PR #735, opened 2026-03-03
Fix invalid DataFrame column name validation before lowercasing.

hongzhi-gao:
Reject None/empty column names in validate_dataframe_for_tsfile and add a unit test to verify the explicit ValueError.


hongzhi-gao:
@ColinLeeo Leave a message to review it, and I’ll merge it directly.

jt2594838 (approved):

PR #745, opened 2026-03-17
Feature/query by row

hongzhi-gao:
## Summary

Add `queryByRow(paths/table, offset, limit)` for both tree and table model. Results are equivalent to “full query then skip first `offset` rows and take at most `limit` rows,” but offset/limit are pushed down so that Chunk/Page-level skipping avoids decoding unnecessary data where possible.

### Tree model
- **API**: `TsFileReader::queryByRow(path_list, offset, limit)` / `TsFileTreeReader::queryByRow(devices, measurements, offset, limit)`.
- **Pushdown**: Single-path: `set_row_range(offset, limit)` on SSI → Chunk/Page skipped by count. Multi-path: offset/limit applied in merge loop; `min_time_hint` used to skip stale Chunks.
- **Tests**: Correctness (no offset/limit, offset only, limit only, offset+limit, boundaries, multi-path merge) + **QueryByRowFasterThanManualNext** (timing: queryByRow faster than full query + manual next, 5% tolerance).

### Table model
- **API**: `TsFileReader::queryByRow(table_name, column_names, offset, limit)`.
- **Pushdown**:  
  - Device: skip whole device when `remaining_offset >= device_row_count` (Dense (in this codebase) means: within one device, every queried column has the same number of rows and the same timestamps.).  
  - SSI: when dense, `set_row_range(offset, limit)` on each column’s SSI → Chunk/Page skip by count.  
  - TsBlock: when sparse or not fully consumed at SSI, offset/limit applied in merge loop.
- **Tests**: Correctness (single/multi device, offset/limit, boundaries, equivalence with manual skip, SSI-level pushdown) + **QueryByRowFasterThanManualNext** (same timing check as tree).

## Review focus

- **Semantics**: queryByRow(offset, limit) matches “full query + skip offset + take limit” (existing equivalence tests).
- **Performance**: New timing tests require queryByRow to be no slower than manual next within 5% (min of 5 runs); confirms pushdown is used in practice.


cpp/src/reader/chunk_reader.cc
```diff
@@ -490,4 +504,82 @@ int ChunkReader::decode_tv_buf_into_tsblock_by_datatype(ByteStream& time_in,
     return ret;
 }
 
+bool ChunkReader::should_skip_page_by_time(int64_t min_time_hint) {
+    if (min_time_hint < 0) {
+        return false;
+    }
```
jt2594838, line 512: Though rare, the timestamp can be negative.
May use the minimum of int64 for this case.

cpp/src/reader/qds_without_timegenerator.cc
```diff
@@ -111,37 +143,111 @@ void QDSWithoutTimeGenerator::close() {
...
+                time_iters_[idx]->next();
+            } else {
+                get_next_tsblock(idx, false);
+            }
+
+            // Apply offset: skip this row.
+            if (remaining_offset_ > 0) {
+                remaining_offset_--;
+                continue;
+            }
```
jt2594838: May skip setting row_record's field if offset > 0.

cpp/src/reader/qds_without_timegenerator.cc
```diff
@@ -199,4 +305,42 @@ int QDSWithoutTimeGenerator::get_next_tsblock(uint32_t index, bool alloc_mem) {
...
+        if (value_iters_[index]) {
+            delete value_iters_[index];
+            value_iters_[index] = nullptr;
+        }
+        if (tsblocks_[index]) {
+            ssi_vec_[index]->destroy();
+            tsblocks_[index] = nullptr;
+        }
+        ret = E_OK;
+    }
```
jt2594838: The failure status cannot be sensed in this branch. Is this safe?

cpp/src/reader/table_query_executor.cc
```diff
@@ -88,6 +88,73 @@ int TableQueryExecutor::query(const std::string& table_name,
...
+    for (size_t i = 0; i < lower_case_column_names.size(); ++i) {
+        column_mapping->add(lower_case_column_names[i], static_cast<int>(i),
+                            *table_schema);
+    }
+    std::vector<common::TSDataType> data_types;
+    data_types.reserve(lower_case_column_names.size());
+    for (size_t i = 0; i < lower_case_column_names.size(); ++i) {
+        auto ind = table_schema->find_column_index(lower_case_column_names[i]);
+        if (ind < 0) {
+            delete time_filter;
```
jt2594838, line 133: Why is time_filter deleted here?

cpp/src/reader/table_query_executor.cc
```diff
@@ -88,6 +88,73 @@ int TableQueryExecutor::query(const std::string& table_name,
     return ret;
 }
 
+int TableQueryExecutor::query(const std::string& table_name,
+                              const std::vector<std::string>& columns,
+                              Filter* time_filter, Filter* id_filter,
```
jt2594838: id_filter -> tag_filter

cpp/src/reader/tsfile_reader.cc
```diff
@@ -116,6 +114,41 @@ int TsFileReader::query(const std::string& table_name,
...
+    return ret;
+}
+
+int TsFileReader::queryByRow(const std::string& table_name,
+                             const std::vector<std::string>& column_names,
+                             int offset, int limit, ResultSet*& result_set) {
+    int ret = E_OK;
+    TsFileMeta* tsfile_meta = tsfile_executor_->get_tsfile_meta();
+    if (tsfile_meta == nullptr) {
+        return E_TSFILE_WRITER_META_ERR;
```
jt2594838: TSFILE_WRITER ?

cpp/src/reader/tsfile_series_scan_iterator.cc
```diff
@@ -36,52 +36,103 @@ void TsFileSeriesScanIterator::destroy() {
     }
 }
 
+bool TsFileSeriesScanIterator::should_skip_chunk_by_time(
+    ChunkMeta* cm, int64_t min_time_hint) {
+    if (min_time_hint < 0 || cm->statistic_ == nullptr) {
```
jt2594838: Beware of negative timestamp

cpp/src/reader/chunk_reader.cc
hongzhi-gao: I changed the min_time_hint sentinel logic to use INT64_MIN instead of checking < 0, so negative timestamps are handled correctly.

cpp/src/reader/qds_without_timegenerator.cc
hongzhi-gao: I updated the flow to skip populating row_record fields when offset is still pending, while still advancing iterators/heap state correctly.

cpp/src/reader/qds_without_timegenerator.cc
hongzhi-gao: I refined this branch to only convert E_NO_MORE_DATA to E_OK. Other failure codes are now preserved and propagated.

cpp/src/reader/table_query_executor.cc
hongzhi-gao: I removed the delete there, since this layer should not own/free time_filter in that path.

cpp/src/reader/table_query_executor.cc
hongzhi-gao: I renamed id_filter to tag_filter in the query interfaces and corresponding usages for consistency.

cpp/src/reader/tsfile_reader.cc
hongzhi-gao: I replaced the reader-side null-meta return code with E_FILE_READ_ERR instead of E_TSFILE_WRITER_META_ERR.

cpp/src/reader/tsfile_series_scan_iterator.cc
hongzhi-gao: I switched to INT64_MIN sentinel handling here as well, so negative timestamps are not treated as an invalid hint.

cpp/src/reader/table_query_executor.cc
jt2594838: Double-check this, instead of removing the delete, more deletes are even added.

jt2594838 (approved):
May also add in C and Python.

hongzhi-gao:
> May also add in C and Python.

Done

PR #746, opened 2026-03-19
enable memory detect.

ColinLeeo:
**Test Configuration**
--------------------------------------------------
  **Write Configuration:**
      Table:         "testTable0"
      TAG columns:   2 (id0, id1) — STRING, PLAIN, UNCOMPRESSED
      FIELD columns: 3 (s0, s1, s2) — INT64, PLAIN, UNCOMPRESSED
      Devices:       10 per flush
      Timestamps:    100 per device per flush
      Rows per flush: 1000 (10 devices × 100 timestamps)
      Flush cycles:  10
      Total rows:    10000

  **Read Configuration:**
      Query:         full table scan (time range [0, INT64_MAX])
      Mode:          row-by-row iteration via TableResultSet::next()
      Sample points: every 2000 rows (5 samples total)

mem_stat

### 1. Write & Flush Behavior
- PAGE_WRITER_OUTPUT_STREAM: allocated during encoding, fully released on flush — produces a sawtooth memory pattern.
- TSFILE_WRITER_META: allocated during flush to serialize ChunkMeta/Statistics into the in-memory index. 
- Accumulates across flushes (~3 KB/flush) and is only released at writer close. This is by design — the full metadata index must be retained until the TsFileMetadata footer is written.
- TSFILE_WRITE_STREAM: the file I/O output buffer, first appears after the initial flush.

### 2. Overall Memory Growth During Write

- Total memory rises steadily due to TSFILE_WRITER_META accumulation (57.6 KB over 20 flushes).
- All other allocations (encoder, compressor, TS2Diff) remain constant after initialization.

### 3. Residual Memory After Writer Close

- 85.7 KB remains after close(), belonging to objects owned by the TsFileTableWriter shared pointer (PageArena, encoders, compressors, statistics).
- These are fully released when the shared pointer goes out of scope — not a leak, just deferred destruction.

### 4. Read Process Memory

- TSFILE_READER grows linearly at ~33.5 bytes/row during row-by-row iteration.
Root cause: TableResultSet::next() allocates string values from a PageArena, which is an arena allocator — it only frees memory in bulk on destroy(), not per-allocation. Previous rows' string memory becomes unreachable but stays in the arena.
- Potential fix: periodically reset the arena when row_record is reset.
TSBLOCK (74.5 KB) is stable — decoded data buffers are reused across pages.
- All reader-specific memory (TSFILE_READER, TSBLOCK, BLOOM_FILTER, DEVICE_*_ITER) is fully released after result set destroy + reader close.

### 5. Residual Memory After Reader Close

- 85.7 KB remaining is entirely from the writer's deferred destruction (same objects as point 3). Reader itself releases all its allocations cleanly.


hongzhi-gao:
Good work!!!

jt2594838 (approved):

PR #747, opened 2026-03-19
cpp tablet new interface

hongzhi-gao:
Add tsfile-cpp tablet new interface

```
    /**
     * @brief Constructs a Tablet object by taking pre-filled columns.
     *
     * For numeric types, `columns[col_index]` should point to a contiguous
     * array of the corresponding C++ type.
     * For string-like types (TEXT/BLOB/STRING), `columns[col_index]` should
     * point to an array of `common::String` with length `row_num`.
     */
    Tablet(uint32_t column_num, uint32_t row_num, void** columns,
           const std::vector<std::string>& column_names,
           const std::vector<common::TSDataType>& data_types);

    /**
     * @brief Set a whole column buffer at once.
     *
     * @param column_index Schema index of the column to set.
     * @param column A pointer to an array with `column_length` elements.
     *               See the constructor doc for the expected element types.
     * @param column_length Number of rows to copy from `column`.
     */
    int set_column(int column_index, void* column, int column_length);
```

### DIFF
--- a/cpp/src/common/tablet.cc
+++ b/cpp/src/common/tablet.cc
@@ -20,6 +20,7 @@
 #include "tablet.h"
 
 #include <cstdlib>
+#include <cstring>
 
 #include "datatype/date_converter.h"
 #include "utils/errno_define.h"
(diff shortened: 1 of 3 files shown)


cpp/src/common/tablet.cc
```diff
@@ -28,6 +29,112 @@ using namespace common;
...
+        return common::E_OK;
+    }
+
+    switch (schema.data_type_) {
+        case common::BOOLEAN:
+            memcpy(col_values.bool_data, column, sizeof(bool) * len);
+            for (uint32_t r = 0; r < len; ++r) {
+                col_notnull_bitmap.clear(r);
+            }
+            break;
```
jt2594838, line 91: Need to directly assign the pointer instead of a memory copy.

The method should be extremely fast but may be unsafe.

cpp/src/common/tablet.cc
```diff
@@ -28,6 +29,112 @@ using namespace common;
...
+    auto& col_values = value_matrix_[column_index];
+    BitMap& col_notnull_bitmap = bitmaps_[column_index];
+
+    if (column == nullptr) {
+        // Treat as all-null.
+        for (uint32_t r = 0; r < len; ++r) {
+            col_notnull_bitmap.set(r);
+        }
+        cur_row_size_ = std::max(cur_row_size_, len);
+        return common::E_OK;
```
jt2594838, line 82: It seems that the bitmap is a null_bitmap.

cpp/src/common/tablet.h
```diff
@@ -162,6 +162,28 @@ class Tablet {
...
+     * For string-like types (TEXT/BLOB/STRING), `columns[col_index]` should
+     * point to an array of `common::String` with length `row_num`.
+     */
+    Tablet(uint32_t column_num, uint32_t row_num, void** columns,
+           const std::vector<std::string>& column_names,
+           const std::vector<common::TSDataType>& data_types);
+
+    /**
+     * @brief Set a whole column buffer at once.
+     *
```
jt2594838, line 179: Mention how cur_row_size is updated after calling this method.

cpp/src/common/tablet.h
```diff
@@ -162,6 +162,28 @@ class Tablet {
...
+     * @param column_index Schema index of the column to set.
+     * @param column A pointer to an array with `column_length` elements.
+     *               See the constructor doc for the expected element types.
+     * @param column_length Number of rows to copy from `column`.
+     */
+    int set_column(int column_index, void* column, int column_length);
+
```
jt2594838, line 186: How about the time column.

PR #748, opened 2026-03-19
Perf: optimize Tablet write with columnar string storage and lazy DeviceID construction (~10x throughput)

ColinLeeo:
Write 200 x 50000 rows: 15039 ms
Throughput: 664938 rows/s

截屏2026-03-20 00 47 02

Write 200 x 50000 rows: 1578 ms
Throughput: 6337140 rows/s

截屏2026-03-20 00 51 04
[collapsed.zip](https://github.com/user-attachments/files/26122101/collapsed.zip)

### DIFF
--- a/cpp/src/common/tablet.cc
+++ b/cpp/src/common/tablet.cc
@@ -245,7 +245,6 @@ int Tablet::set_column_values(uint32_t schema_index, const void* data,
             break;
         default:
             return E_TYPE_NOT_SUPPORTED;
-            ;
     }
 
     std::memcpy(dst, data, count * elem_size);
@@ -278,24 +277,16 @@ int Tablet::set_column_string_values(uint32_t schema_index,
         return E_INVALID_ARG;
     }
 
-    int32_t base = offsets[0];
-    uint32_t total_bytes = static_cast<uint32_t>(offsets[count] - base);
+    uint32_t total_bytes = static_cast<uint32_t>(offsets[count]);
     if (total_bytes > sc->buf_capacity) {
         sc->buf_capacity = total_bytes;
         sc->buffer = (char*)mem_realloc(sc->buffer, sc->buf_capacity);
     }
 
     if (total_bytes > 0) {
-        std::memcpy(sc->buffer, data + base, total_bytes);
+        std::memcpy(sc->buffer, data, total_bytes);
     }
-
-    if (base == 0) {
-        std::memcpy(sc->offsets, offsets, (count + 1) * sizeof(int32_t));
-    } else {
-        for (uint32_t i = 0; i <= count; i++) {
-            sc->offsets[i] = offsets[i] - base;
-        }
-    }
+    std::memcpy(sc->offsets, offsets, (count + 1) * sizeof(int32_t));
     sc->buf_used = total_bytes;
 
     if (bitmap == nullptr) {
@@ -342,6 +333,8 @@ void* Tablet::get_value(int row_index, uint32_t schema_index,
             double* double_values = column_values.double_data;
             return &double_values[row_index];
         }
+        case TEXT:
+        case BLOB:
         case STRING: {
             return &column_values.string_col->get_string_view(row_index);
         }
--- a/cpp/src/common/tablet.cc
+++ b/cpp/src/common/tablet.cc
@@ -279,8 +279,7 @@ int Tablet::set_column_string_repeated(uint32_t schema_index, const char* str,
         uint32_t filled = 1;
         while (filled < count) {
             uint32_t batch = std::min(filled, count - filled);
-            memcpy(sc->buffer + filled * str_len, sc->buffer,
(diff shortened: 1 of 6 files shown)


hongzhi-gao:
LGTM

cpp/src/common/tablet.cc
```diff
@@ -98,10 +98,9 @@ int Tablet::init() {
             case BLOB:
             case TEXT:
             case STRING: {
-                value_matrix_[c].string_data =
-                    static_cast<common::String*>(common::mem_alloc(
-                        sizeof(String) * max_row_num_, common::MOD_TABLET));
-                if (value_matrix_[c].string_data == nullptr) return E_OOM;
+                auto* sc = new StringColumn();
+                sc->init(max_row_num_, max_row_num_ * 32);
+                value_matrix_[c].string_col = sc;
```
jt2594838: Use mem_alloc?

cpp/src/common/tablet.cc
```diff
@@ -444,6 +444,57 @@ void Tablet::set_column_categories(
...
+            if (boundary[i >> 6] & (1ULL << (i & 63))) continue;
+            uint32_t len_a = off[i] - off[i - 1];
+            uint32_t len_b = off[i + 1] - off[i];
+            if (len_a != len_b ||
+                (len_a > 0 &&
+                 memcmp(buf + off[i - 1], buf + off[i], len_a) != 0)) {
+                boundary[i >> 6] |= (1ULL << (i & 63));
+            }
+        }
+    }
```
jt2594838: May traverse tag columns in reversed order, because we tend to organize tags from big (like country) to small (like street).
You are more likely to find differences between small tags within the same TsFile or write batch.

cpp/src/common/tablet.cc
```diff
@@ -444,6 +444,57 @@ void Tablet::set_column_categories(
...
+        const uint32_t* off = sc.offsets;
+        const char* buf = sc.buffer;
+        for (uint32_t i = 1; i < row_count; i++) {
+            if (boundary[i >> 6] & (1ULL << (i & 63))) continue;
+            uint32_t len_a = off[i] - off[i - 1];
+            uint32_t len_b = off[i + 1] - off[i];
+            if (len_a != len_b ||
+                (len_a > 0 &&
+                 memcmp(buf + off[i - 1], buf + off[i], len_a) != 0)) {
+                boundary[i >> 6] |= (1ULL << (i & 63));
```
jt2594838, line 544: If the number of boundaries reaches the number of rows, may break.

cpp/src/common/tablet.h
```diff
@@ -46,14 +46,78 @@ class TabletColIterator;
...
+            // Grow buffer if needed
+            if (buf_used + len > buf_capacity) {
+                buf_capacity = buf_capacity * 2 + len;
+                buffer = (char*)common::mem_realloc(buffer, buf_capacity);
+            }
+            memcpy(buffer + buf_used, data, len);
+            offsets[row] = buf_used;
+            offsets[row + 1] = buf_used + len;
+            buf_used += len;
+        }
```
jt2594838, line 93: If data equals the value of the previous row, may simply use the same offsets and avoid a memory copy.
However, if the memory comparison is too often, but the memory copy is not avoided, we should stop comparing them.

cpp/src/common/tablet.cc
ColinLeeo: fixed

cpp/src/common/tablet.cc
ColinLeeo: Fixed

cpp/src/common/tablet.h
ColinLeeo: The Arrow-style offsets buffer requires that each row be appended in order. If we allow two rows to share the same segment of data (i.e., offsets[row] == offsets[row-1] and offsets[row+1] == offsets[row]), then we must handle this kind of “zero-length range” either in the offset normalization path of set_column_string_values or add special handling during writing.

This would introduce additional branching and complexity into all downstream logic (including Arrow conversion).

cpp/src/common/tablet.cc
ColinLeeo: fixed

cpp/src/common/tablet.cc
```diff
@@ -150,7 +155,9 @@ void Tablet::destroy() {
                 case BLOB:
                 case TEXT:
                 case STRING:
-                    common::mem_free(value_matrix_[c].string_data);
+                    value_matrix_[c].string_col->destroy();
+                    value_matrix_[c].string_col->~StringColumn();
+                    common::mem_free(value_matrix_[c].string_col);
```
jt2594838, line 159: It is a bit weird to call both destroy and destruction.
Why is destory not included in destruction?

jt2594838 (approved):

PR #753, opened 2026-03-26
Add tsblock tsfile writer

shuwenwei:


JackieTien97 (approved):

java/tsfile/src/main/java/org/apache/tsfile/write/v4/TableTsBlock2TsFileWriter.java
```diff
@@ -0,0 +1,266 @@
...
+     * Write a range of rows into chunk group.
+     *
+     * @param startRowIndex inclusive
+     * @param endRowIndex exclusive
+     */
+    public int write(Column timeColumn, Column[] valueColumns, int startRowIndex, int endRowIndex)
+        throws WriteProcessException {
+      int pointCount = 0;
+      for (int rowIndex = startRowIndex; rowIndex < endRowIndex; rowIndex++) {
+        long time = timeColumn.getLong(rowIndex);
```
JackieTien97, line 223: it may be better to check timeColumn.isNull, you can throw exception if so.

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #764, opened 2026-04-01
Fix/tree query and rle decoder

ColinLeeo:
This pull request introduces several important bug fixes and improvements to the RLE (Run-Length Encoding) decoding logic, especially for handling large run counts and memory management, as well as fixes for device path handling and table lookup in the TsFile tree reader. It also adds comprehensive regression tests to prevent recurrence of these issues.

### RLE Decoder Improvements and Bug Fixes

* Updated `Int32RleDecoder` to correctly read multi-byte LEB128 varint headers for RLE run counts, fixing decoding errors for large runs (e.g., run_count ≥ 64). The decoder now uses `read_var_uint` for the header and properly distinguishes between RLE and bit-packing modes. (`cpp/src/encoding/int32_rle_decoder.h`)
* Fixed memory management in `Int32RleDecoder::reset()` to use `mem_free` instead of `delete[]` for freeing buffers allocated with `mem_alloc`, preventing crashes and undefined behavior. (`cpp/src/encoding/int32_rle_decoder.h`)
* Added and updated regression tests for RLE decoding, covering large run counts, multiple consecutive runs, and memory management after reset. (`cpp/test/encoding/int32_rle_codec_test.cc`)

### Device Path Parsing and Table Lookup Fixes

* Fixed `StringArrayDeviceID` construction to correctly split device ID strings, ensuring accurate extraction of table names and device names for deep device paths (e.g., "root.sensors.TH"). (`cpp/src/common/device_id.cc`) [[1]](diffhunk://#diff-0412bd32d8c9375111d4a2686152e750ec043aa562e4ded96d088ed465132d94L61-R61) [[2]](diffhunk://#diff-0412bd32d8c9375111d4a2686152e750ec043aa562e4ded96d088ed465132d94L212-R215)
* Improved `TsFileIOReader::load_device_index_entry` to use `find()` instead of `operator[]` when looking up table metadata, avoiding unintended insertion of null entries and returning proper error codes when a table or device does not exist. (`cpp/src/file/tsfile_io_reader.cc`)

### Test Enhancements

* Added regression tests in `TsFileTreeReaderTest` for:
  - Querying devices with deep dot-separated paths, ensuring correct table and device name extraction.
  - Handling missing measurements gracefully without crashing, verifying that the correct error code is returned instead of a segmentation fault. (`cpp/test/reader/tree_view/tsfile_reader_tree_test.cc`)

These changes collectively improve the robustness and correctness of RLE decoding and device/table handling in the codebase.

### DIFF
--- a/cpp/src/encoding/int32_rle_decoder.h
+++ b/cpp/src/encoding/int32_rle_decoder.h
@@ -139,8 +139,9 @@ class Int32RleDecoder : public Decoder {
             }
         }
         --current_count_;
-        result = is_rle_run_ ? rle_value_
-                             : current_buffer_[bitpacking_num_ - current_count_ - 1];
+        result = is_rle_run_
+                     ? rle_value_
+                     : current_buffer_[bitpacking_num_ - current_count_ - 1];
         if (!has_next_package()) {
             is_length_and_bitwidth_readed_ = false;
         }
--- a/cpp/src/encoding/int32_rle_decoder.h
+++ b/cpp/src/encoding/int32_rle_decoder.h
@@ -64,13 +64,14 @@ class Int32RleDecoder : public Decoder {
     }
     int read_boolean(bool& ret_value, common::ByteStream& in) override {
         int32_t bool_value;
-        read_int32(bool_value, in);
-        ret_value = bool_value == 0 ? false : true;
-        return common::E_OK;
+        int ret = read_int32(bool_value, in);
+        if (ret == common::E_OK) {
+            ret_value = bool_value != 0;
+        }
+        return ret;
     }
     int read_int32(int32_t& ret_value, common::ByteStream& in) override {
-        ret_value = static_cast<int32_t>(read_int(in));
-        return common::E_OK;
+        return read_int(ret_value, in);
     }
     int read_int64(int64_t& ret_value, common::ByteStream& in) override {
         return common::E_TYPE_NOT_MATCH;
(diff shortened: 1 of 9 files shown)


cpp/src/encoding/int32_rle_decoder.h
```diff
@@ -125,8 +133,40 @@ class Int32RleDecoder : public Decoder {
...
+            sizeof(int32_t) * run_length, common::MOD_DECODER_OBJ));
+        if (IS_NULL(current_buffer_)) {
+            return common::E_OOM;
+        }
+        for (int i = 0; i < run_length; i++) {
+            current_buffer_[i] = value;
+        }
+        current_count_ = run_length;
+        bitpacking_num_ = run_length;
+        return ret;
```
jt2594838: Is it possible to set a flag for RLE run, and we just store the repeated value instead of using current_buffer_?

jt2594838 (approved):

cpp/src/encoding/int32_rle_decoder.h
```diff
@@ -109,24 +115,57 @@ class Int32RleDecoder : public Decoder {
...
-                    common::SerializationUtil::read_ui8(header, byte_cache_))) {
+            if (RET_FAIL(common::SerializationUtil::read_var_uint(
+                    header_value, byte_cache_))) {
                 return ret;
             }
-            call_read_bit_packing_buffer(header);
+            if (header_value & 1) {
+                call_read_bit_packing_buffer(header_value);
+            } else {
+                call_read_rle_run(header_value);
```
hongzhi-gao: Please check these return values and propagate failures.

hongzhi-gao (approved):
LGTM

PR #765, opened 2026-04-02
Support TsfileDataFrame

ycycse:
This PR introduce TsFileDataFrame， which can read multi tsfile for model training usage. Just like use dataframe.

https://apache-iotdb-project.feishu.cn/docx/SenJdxlbuoUS5Uxmq7jcOUzdnob?from=from_copylink

  ## Build Wheel

  Prerequisite: build the C++ library first, since the Python package depends on the shared library from `cpp/target/build`.

  From the repo root:

  ```bash
  mvn -P with-cpp clean install -DskipTests

  cd python
  python setup.py build_ext --inplace
  python setup.py bdist_wheel

  pip install dist/*.whl

```

  ## Validation
```
  Basic dataset regression tests:

  pytest -q python/tests/test_tsfile_dataset.py
```

  ## Basic Usage

  ### Open one TsFile
```python
  from tsfile import TsFileDataFrame

  tsdf = TsFileDataFrame("/path/to/file.tsfile")
  print(tsdf)
```
  ### Open a shard directory

  TsFileDataFrame will recursively collect all .tsfile files under the directory.
```
  from tsfile import TsFileDataFrame

  tsdf = TsFileDataFrame("/path/to/dataset_dir")
  print(len(tsdf))
  print(tsdf.list_timeseries()[:5])
```
  ### Filter by metadata
```
  filter_df = tsdf[tsdf['filed'] == 'weather']
  print(filter_df)
```
  ### Access one logical series

  By index:
```
  series = tsdf[0]
  print(series)
  print(series.name)
  print(series.stats)
  print(series[0])
  print(series[:10])
```
  By logical series path:
```
  series = tsdf["weather.device_a.temperature"]
  print(series.timestamps)
```

  ### Build subset views
```
  subset = tsdf[:10]
  print(subset)

  subset2 = tsdf[[0, 3, 5]]
  print(subset2.list_timeseries())
```
  ### Time-aligned multi-series query
```
  aligned = tsdf.loc[0:1000, [0, 1]]
  print(aligned)
  print(aligned.timestamps)
  print(aligned.values)
  print(aligned.series_names)
```
  Or use logical series names:
```
  aligned = tsdf.loc[
      0:1000,
      ["weather.device_a.temperature", "weather.device_a.humidity"],
  ]
```


python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+
+        Args:
+            max_rows: Maximum rows to display. None for all rows.
+        """
+        n_rows, n_cols = self.values.shape
+        if n_rows == 0:
+            print(f"AlignedTimeseries(0 rows, {n_cols} series)")
+            return
+        ts_strs, ts_width, col_widths, val_strs = self._build_display()
+        print(self._format_rows(ts_strs, ts_width, col_widths, val_strs, max_rows))
```
jt2594838: Pass max_rows to _build_display?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+            col_widths.append(w)
+
+        return ts_strs, ts_width, col_widths, val_strs
+
+    def _format_rows(self, ts_strs, ts_width, col_widths, val_strs, max_rows):
+        """Format rows with optional truncation."""
+        n_rows = len(ts_strs)
+        n_cols = len(col_widths)
+
+        header_parts = ['timestamp'.rjust(ts_width)]
```
jt2594838: The column is a datetime string; it is improper to call it `timestamp`.
May use `time` or `datetime`.

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+        """Pre-compute string representations for display."""
+        n_rows, n_cols = self.values.shape
+        ts_strs = [_format_timestamp(int(t)) for t in self.timestamps]
+        ts_width = max((len(s) for s in ts_strs), default=0)
+        ts_width = max(ts_width, len('timestamp'))
+
+        col_widths = []
+        val_strs = []
+        for col_idx in range(n_cols):
+            col_name = self.series_names[col_idx] if col_idx < len(self.series_names) else f'col_{col_idx}'
```
jt2594838: In what case will col_xxx be used?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+    def _build_display(self):
+        """Pre-compute string representations for display."""
+        n_rows, n_cols = self.values.shape
+        ts_strs = [_format_timestamp(int(t)) for t in self.timestamps]
+        ts_width = max((len(s) for s in ts_strs), default=0)
+        ts_width = max(ts_width, len('timestamp'))
+
+        col_widths = []
+        val_strs = []
+        for col_idx in range(n_cols):
```
jt2594838: Is it possible to concat timestamps and values, then use ndarry's print method if any?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+        # Merge from multiple readers, sort by timestamp, deduplicate
+        merged_ts = np.concatenate(all_ts)
+        merged_vals = np.concatenate(all_vals)
+        sort_idx = np.argsort(merged_ts, kind='mergesort')
+        merged_ts = merged_ts[sort_idx]
+        merged_vals = merged_vals[sort_idx]
+
+        # Deduplicate by timestamp (keep first occurrence)
+        _, unique_idx = np.unique(merged_ts, return_index=True)
+        return merged_ts[unique_idx], merged_vals[unique_idx]
```
jt2594838: Why keep the first occurrence?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+            else:
+                raise TypeError(f"Series specifier must be int or str, got {type(s)}")
+
+        return start_time, end_time, series_names
+
+    def _query_aligned(self, start_time: int, end_time: int, series_names: List[str]):
+        """Query and align multiple series using batched Arrow reads."""
+        # Group series by (reader_id, table_name, tag_tuple) for batch queries.
+        # Each group can be fetched with a single query_table_batch call.
+        groups = defaultdict(list)  # key -> [(col_idx, field_name, series_name, reader, info)]
```
jt2594838: What is the relation between field_name and series_name?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+            info = entries[0][4]
+            field_columns = list(dict.fromkeys(e[1] for e in entries))  # dedupe, keep order
+
+            ts_arr, field_vals = reader.read_multi_series_by_time_range(
+                info['table_name'], field_columns,
+                info['tag_columns'], info['tag_values'],
+                start_time, end_time,
+            )
+
+            for _col_idx, field_name, name, _, _ in entries:
```
jt2594838: _col_idx is not used?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+                    prev_ts, prev_val = series_data[name]
+                    merged_ts = np.concatenate([prev_ts, ts_arr])
+                    merged_val = np.concatenate([prev_val, field_vals[field_name]])
+                    sort_idx = np.argsort(merged_ts, kind='mergesort')
+                    merged_ts = merged_ts[sort_idx]
+                    merged_val = merged_val[sort_idx]
+                    _, unique_idx = np.unique(merged_ts, return_index=True)
+                    series_data[name] = (merged_ts[unique_idx], merged_val[unique_idx])
+                else:
+                    series_data[name] = (ts_arr, field_vals[field_name])
```
jt2594838: For n files with k points each, there will be n merges, each merging (k+k, 2k+k, 3k+k, ...) points?
The overall complexity seems too high.

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+
+        # Collect unique timestamps using numpy
+        all_ts_arrays = [data[0] for data in series_data.values() if len(data[0]) > 0]
+        if not all_ts_arrays:
+            return (np.array([], dtype=np.int64),
+                    np.array([]).reshape(0, len(series_names)))
+
+        sorted_timestamps = np.unique(np.concatenate(all_ts_arrays))
```
jt2594838: Where is it sorted?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+        # Expand directories: collect all .tsfile files within each directory
+        expanded = []
+        for p in paths:
+            if os.path.isdir(p):
+                tsfiles = sorted(
+                    os.path.join(root, f)
+                    for root, _, files in os.walk(p)
+                    for f in files
+                    if f.endswith('.tsfile')
+                )
```
jt2594838: Will os.walk(p) recursively search all subdirs?

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+        """Attribute-style access to the loc indexer."""
+        return _LocIndexer(self)
+
+    def _collect_tag_columns(self) -> List[str]:
+        """Return ordered list of all unique tag column names across all series."""
+        seen = {}
+        for name in self._series_list:
+            for col in self._merged_info[name]['tag_columns']:
+                if col not in seen:
+                    seen[col] = True
```
jt2594838: May store a map table_name -> tag_columns to reduce repeated computation.

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+            raise ValueError("No tables found in TsFile")
+
+        self.series_paths = []
+        table_names = list(table_schemas.keys())
+
+        # Progress tracking
+        total_rows = 0
+
+        for ti, table_name in enumerate(table_names):
+            table_schema = self._reader.get_table_schema(table_name)
```
jt2594838: Why not use table_schemas?

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+
+            tag_columns = []
+            field_columns = []
+            for col_schema in table_schema.get_columns():
+                col_name = col_schema.get_column_name()
+                col_category = col_schema.get_category()
+                if col_name.lower() == 'time':
+                    continue
```
jt2594838: Also check column category?

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+
+            # Query TAG columns + first FIELD column to discover groups and timestamps
+            query_cols = tag_columns + [field_columns[0]]
+
+            time_arrays = []
+            tag_arrays = {tc: [] for tc in tag_columns}
+
+            with self._reader.query_table_batch(
+                table_name, query_cols, batch_size=65536
+            ) as rs:
```
jt2594838: A TsFile may not actually contain a column's data even if it is in the schema.
Double-check if timestamps can be read in this scenario.

jt2594838: Storing tags repeatedly is a great challenge to memory.
May consider getting all DeviceIds first, and conclude tag values from them.

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+                        col_mask = pc.equal(arrow_table.column(tag_col), tag_val)
+                        mask = col_mask if mask is None else pc.and_(mask, col_mask)
+                    arrow_table = arrow_table.filter(mask)
+
+                if arrow_table.num_rows > 0:
+                    ts_list.append(arrow_table.column('time').to_numpy())
+                    for fc in field_columns:
+                        field_lists[fc].append(
+                            arrow_table.column(fc).to_numpy().astype(np.float64)
+                        )
```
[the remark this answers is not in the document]
jt2594838: +1

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+        """Pre-load series data into memory cache.
+
+        Args:
+            series_path: Time series path.
+        """
+        if series_path not in self.series_info:
+            raise ValueError(f"Series not found: {series_path}")
+        if series_path not in self._series_data_cache:
+            data = self.read_series(series_path)
+            self._series_data_cache[series_path] = np.array(data, dtype=np.float32)
```
jt2594838: The method is not used? And there does not seem to be any memory control.

JackieTien97: `read_series` returns float64 values (via `_read_arrow` which casts to `np.float64`), but `cache_series_data` stores them as float32. When `read_series` is later called on a cached series, it returns float32 values silently, which can lose precision.

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+                    tag_tuples = [
+                        tuple(all_tags[tc][i] for tc in tag_columns)
+                        for i in range(n)
+                    ]
+                    unique_tuples = list(dict.fromkeys(tag_tuples))
+                    for ut in unique_tuples:
+                        mask = np.array([t == ut for t in tag_tuples], dtype=bool)
+                        self._register_tag_group(
+                            table_name, tag_columns, ut,
+                            field_columns, all_times[mask]
```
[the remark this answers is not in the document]
JackieTien97: For tables with multiple tag columns, grouping is done by:
1. Building a Python list of tuples (line 174-177): O(n)
2. For each unique tuple, scanning the entire list to build a boolean mask (line 180): O(n * k) where k = number of unique groups

For a table with 1M rows and 1000 tag groups, this becomes ~1 billion comparisons in pure Python.

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+        # Parse series names
+        if isinstance(series_spec, (str, int)):
+            series_spec = [series_spec]
+
+        series_names = []
+        for s in series_spec:
+            if isinstance(s, (int, np.integer)):
+                idx = int(s)
+                if idx < 0 or idx >= len(self._df._series_list):
+                    raise IndexError(f"Series index {idx} out of range")
```
JackieTien97: Negative indices are not normalized (no `idx = length + idx` like other `__getitem__` methods in this file). A user passing `df.loc[:, [-1]]` would get an IndexError instead of the last series.

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+                'min_time': info['min_time'],
+                'max_time': info['max_time'],
+                'count': info['length'],
+            }
+        else:
+            all_ts = []
+            for reader, info in entries:
+                all_ts.append(reader._timestamps_cache[series_path])
+            merged = np.unique(np.concatenate(all_ts))
+            merged.sort()
```
JackieTien97: `np.unique` already returns a sorted array. The `.sort()` call is redundant.

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+        header_parts = ['timestamp'.rjust(ts_width)]
+        for col_idx in range(n_cols):
+            col_name = self.series_names[col_idx] if col_idx < len(self.series_names) else f'col_{col_idx}'
+            header_parts.append(col_name.rjust(col_widths[col_idx]))
+        lines = ['  '.join(header_parts)]
+
+        if max_rows is None or n_rows <= max_rows:
+            show_rows = list(range(n_rows))
+        else:
+            show_rows = list(range(max_rows))
```
JackieTien97: When `max_rows` is exceeded, only the first `max_rows` rows are shown. In contrast, `TsFileDataFrame._format_table` shows head + "..." + tail. This inconsistency may confuse users expecting similar behavior from both display methods.

python/tsfile/__init__.py
```diff
@@ -40,4 +40,5 @@
 from .tsfile_writer import TsFileWriterPy as TsFileWriter
 from .tsfile_py_cpp import get_tsfile_config, set_tsfile_config
 from .tsfile_table_writer import TsFileTableWriter
-from .utils import to_dataframe, dataframe_to_tsfile
\ No newline at end of file
+from .utils import to_dataframe, dataframe_to_tsfile
+from .tsfile_dataframe import TsFileDataFrame, Timeseries, AlignedTimeseries
```
JackieTien97: The file ends without a newline character (`\ No newline at end of file`).

python/tsfile/tsfile_dataframe.py
```diff
@@ -0,0 +1,859 @@
...
+    def close(self):
+        """Close all underlying readers.
+
+        No-op for subset views (they don't own readers).
+        """
+        if self._is_view:
+            return
+        for reader in self._readers.values():
+            reader.close()
+        self._readers.clear()
```
JackieTien97: After `close()`, `_readers` is cleared but `_series_map` still holds references to closed readers. Any subsequent data access (e.g., `tsdf[0][0]`) will attempt to read from a closed reader, producing an unclear error.

**Recommendation**: Either invalidate `_series_map` too, or set a `_closed` flag and check it in data-access paths.

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def _cache_metadata(self):
+        """Cache metadata from the TsFile."""
+        try:
+            self._cache_metadata_table_model()
+        except Exception as e:
+            raise ValueError(
+                f"Failed to read TsFile metadata. "
+                f"Please ensure the TsFile is valid and readable. Error: {e}"
+            )
```
JackieTien97: The original traceback is lost. Use `raise ValueError(...) from e` to preserve the exception chain for debugging.

JackieTien97 (changes requested):

python/tsfile/tsfile_series_reader.py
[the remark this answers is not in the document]
ycycse: Done. The implementation has since been refactored into python/tsfile/dataset/reader.py. The dataset reader now only includes numeric FIELD columns during discovery, and non-numeric fields are excluded from the dataset path. The current read path also keeps an explicit numeric-compatibility check when converting Arrow arrays.

python/tsfile/tsfile_series_reader.py
```diff
@@ -0,0 +1,440 @@
...
+        """
+        if series_path not in self.series_info:
+            raise ValueError(f"Series not found: {series_path}")
+
+        if series_path in self._series_data_cache:
+            return self._series_data_cache[series_path][start:end].tolist()
+
+        info = self.series_info[series_path]
+        timestamps = self._timestamps_cache[series_path]
+
```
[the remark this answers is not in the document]
ycycse: done

python/tsfile/tsfile_series_reader.py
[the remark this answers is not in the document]
ycycse: Done. The implementation has moved to python/tsfile/dataset/reader.py. The pure-Python tuple/mask grouping path was replaced with a vectorized NumPy grouping implementation based on structured arrays and np.unique(..., return_inverse=True).

python/tsfile/tsfile_dataframe.py
ycycse: done. The implementation has moved to python/tsfile/dataset/formatting.py and python/tsfile/dataset/timeseries.py. Now all the formatting use general display logic.

python/tsfile/tsfile_dataframe.py
ycycse: Done. The implementation has moved to python/tsfile/dataset/formatting.py. The aligned display header now uses time instead of timestamp.

python/tsfile/tsfile_dataframe.py
ycycse: The implementation has moved to python/tsfile/dataset/formatting.py. This is now only a defensive fallback for mismatched metadata during formatting. In normal public paths it should not be used, since series_names and value columns are expected to match.

python/tsfile/tsfile_dataframe.py
ycycse: The implementation has moved to python/tsfile/dataset/formatting.py. I kept the custom formatter intentionally, because it gives us stable head/tail truncation, NaN rendering, and consistent column formatting, which ndarray printing does not provide well for this API.

python/tsfile/tsfile_dataframe.py
ycycse: Done. The implementation has moved to python/tsfile/dataset/merge.py. The current merge policy rejects duplicate timestamps across shards instead of keeping the first occurrence.

python/tsfile/tsfile_dataframe.py
ycycse: The implementation has moved to python/tsfile/dataset/metadata.py and python/tsfile/dataset/dataframe.py. In the current model, field_name is the last component of the logical series name. A series name is built as table + tag values + field.

python/tsfile/tsfile_dataframe.py
ycycse: Obsolete after the refactor. The implementation has moved to python/tsfile/dataset/dataframe.py, and that old code path no longer exists.

python/tsfile/tsfile_dataframe.py
ycycse: Addressed. The implementation has moved to python/tsfile/dataset/merge.py and python/tsfile/dataset/dataframe.py. The current merge path uses dedicated merge helpers with a fast concatenate path for non-overlapping ranges and a k-way merge.

python/tsfile/tsfile_dataframe.py
ycycse: np.unique

python/tsfile/tsfile_dataframe.py
ycycse: Sure

python/tsfile/tsfile_dataframe.py
ycycse: done. The implementation has moved to python/tsfile/dataset/dataframe.py and python/tsfile/dataset/metadata.py. Table-level schema metadata is now shared via table entries instead of being rebuilt repeatedly per series.

python/tsfile/tsfile_series_reader.py
ycycse: done.

python/tsfile/tsfile_series_reader.py
ycycse: done

python/tsfile/tsfile_series_reader.py
ycycse: Done. The implementation has moved to python/tsfile/dataset/metadata.py, python/tsfile/dataset/reader.py, and python/tsfile/dataset/dataframe.py. Shared metadata is now centered on table/device entries, and logical series are represented by lightweight refs instead of repeated per-series tag dictionaries.

python/tsfile/tsfile_series_reader.py
ycycse: It's an old method. We don't need it now.

ycycse: Done. The old cached-series path was removed during the refactor, so this precision inconsistency no longer exists.

python/tsfile/tsfile_dataframe.py
ycycse: Done. The implementation has moved to python/tsfile/dataset/dataframe.py. Negative indices in .loc series selection are normalized now.

python/tsfile/tsfile_dataframe.py
ycycse: done

python/tsfile/tsfile_dataframe.py
ycycse: done. The aligned display now uses the same head + ... + tail truncation style.

python/tsfile/dataset/metadata.py
```diff
@@ -0,0 +1,224 @@
...
+        if field_name not in self._field_index_by_name:
+            raise ValueError(f"Field not found in table '{self.table_name}': {field_name}")
+        return self._field_index_by_name[field_name]
+
+
+@dataclass(slots=True)
+class DeviceEntry:
+    """One logical device identified by table name + ordered tag values."""
+
+    table_id: int
```
jt2594838: doc says table name, but it is table_id.

Where is the id associated?

python/tsfile/dataset/metadata.py
```diff
@@ -0,0 +1,224 @@
...
+        self.table_entries.append(
+            TableEntry(
+                table_name=table_name,
+                tag_columns=tuple(tag_columns),
+                tag_types=tuple(tag_types),
+                field_columns=tuple(field_columns),
+            )
+        )
+        self.table_id_by_name[table_name] = table_id
+        return table_id
```
jt2594838, line 93: Is the caller responsible for checking the existence of the table?

Notice that the same table may have different columns in different TsFiles, which should be merged.

jt2594838 (approved):

python/tsfile/dataset/metadata.py
ycycse: That's right, I missed that. In current impentation, same table across TsFiles are assumed to have the same schema which need some time to support field-union merge. Since we have validation to reject the situaion, I think we can note it down and follow up it with a separate PR?

python/tsfile/dataset/metadata.py
ycycse: It's a typo, I will use table_id in doc instread.

python/tsfile/dataset/metadata.py
jt2594838: Ok, I have no further questions then.

python/tsfile/tsfile_dataframe.py
ycycse: done. we have a _closed flag now.

PR #767, opened 2026-04-03
Feat/c python timeseries metadata

hongzhi-gao:
## Summary

Adds C-wrapper support for listing devices and reading per-device timeseries metadata, with Python `TsFileReader` bindings. Device identity is a single `DeviceID` (path, table name, segments). Chunk statistics are exposed as a tagged layout: common header `TsFileStatisticBase` (`has_statistic`, `TSDataType type`, `row_count`, `start_time`, `end_time`) plus a `TimeseriesStatisticUnion` arm selected by `base.type`. Numeric arms expose `sum` (boolean sum of true values as `double`, integer sum as `double`, float/double sum as `double`), ranges, and first/last where applicable; STRING exposes lexicographic min/max and time-ordered first/last strings; TEXT exposes first/last strings only (no min/max).

### C API

**Types:** `TsFileStatisticBase`, `TsFileBoolStatistic`, `TsFileIntStatistic`, `TsFileFloatStatistic`, `TsFileStringStatistic`, `TsFileTextStatistic`, `TimeseriesStatisticUnion`, `TimeseriesStatistic`, `TimeseriesMetadata`, `DeviceID`, `DeviceTimeseriesMetadataEntry` (embeds `device` only), `DeviceTimeseriesMetadataMap`.

**Device listing:** `tsfile_reader_get_all_devices` → `DeviceID*`; free with `tsfile_free_device_id_array`. For a single struct, clear heap fields with `tsfile_device_id_free_contents`.

**Metadata:** `tsfile_reader_get_timeseries_metadata_all` (whole file); `tsfile_reader_get_timeseries_metadata_for_devices` (`length == 0` → empty map; `devices == NULL && length > 0` → invalid arg; when `length > 0`, each entry’s `path` is required). Free maps with `tsfile_free_device_timeseries_metadata_map` only (also frees nested device strings and statistic heap strings).

Read common statistic fields via `tsfile_statistic_base(&statistic)`; read `sum` and type-specific min/max/first/last via `statistic.u.bool_s`, `.int_s`, `.float_s`, `.string_s`, or `.text_s` according to `base.type`.

### Python

- `get_all_devices()` → `list[DeviceID]` (`tsfile.schema.DeviceID`: path, table name, segments).
- `get_timeseries_metadata(device_ids=None)` → `dict[str, DeviceTimeseriesMetadataGroup]`: `None` = all devices, `[]` = empty map, non-empty list filters by device path. Elements may be `DeviceID` (uses `.path`) or any object accepted as path via `str(...)`.

Statistic types map to `TimeseriesStatistic` plus subclasses `IntTimeseriesStatistic`, `FloatTimeseriesStatistic`, `BoolTimeseriesStatistic`, `StringTimeseriesStatistic`, and `TextTimeseriesStatistic` (with `sum` only on the numeric/boolean subclasses).

### Tests

- `cpp/test/cwrapper/cwrapper_metadata_test.cc`
- `python/tests/test_reader_metadata.py`

### DIFF
--- a/cpp/src/cwrapper/tsfile_cwrapper.h
+++ b/cpp/src/cwrapper/tsfile_cwrapper.h
@@ -146,7 +146,8 @@ typedef struct TimeseriesStatistic {
     bool first_bool;
     bool last_bool;
 
-    /** STRING: min/max lexicographic; TEXT: first/last only (min/max unused). */
+    /** STRING: min/max lexicographic; TEXT: first/last only (min/max unused).
+     */
     bool str_ext_valid;
     char* str_min;
     char* str_max;
--- a/cpp/src/cwrapper/tsfile_cwrapper.h
+++ b/cpp/src/cwrapper/tsfile_cwrapper.h
@@ -165,8 +165,29 @@ typedef struct TimeseriesMetadata {
     TimeseriesStatistic statistic;
 } TimeseriesMetadata;
 
+/**
+ * @brief Device identity fields from IDeviceID (path, table name, segments).
+ * Allocated by tsfile_reader_get_all_device_details; freed by
+ * tsfile_free_device_details_array.
+ */
+typedef struct TsDeviceDetails {
+    char* path;
+    char* table_name;
+    uint32_t segment_count;
+    char** segments;
+} TsDeviceDetails;
+
+/**
+ * @brief One device's timeseries metadata list plus structured device fields.
+ *
+ * device_table_name / device_segments are malloc'd; freed by
+ * tsfile_free_device_timeseries_metadata_map (do not free individually).
+ */
 typedef struct DeviceTimeseriesMetadataEntry {
     DeviceID device;
+    char* device_table_name;
+    uint32_t device_segment_count;
+    char** device_segments;
     TimeseriesMetadata* timeseries;
     uint32_t timeseries_count;
 } DeviceTimeseriesMetadataEntry;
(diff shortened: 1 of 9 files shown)


cpp/src/cwrapper/tsfile_cwrapper.h
```diff
@@ -104,6 +104,97 @@ typedef struct device_schema {
...
+    bool last_bool;
+
+    /** STRING: min/max lexicographic; TEXT: first/last only (min/max unused).
+     */
+    bool str_ext_valid;
+    char* str_min;
+    char* str_max;
+    char* str_first;
+    char* str_last;
+} TimeseriesStatistic;
```
jt2594838, line 183: typedef struct BaseStatistics {
   TSDataType type;
    int32_t row_count;
    int64_t start_time;
    int64_t end_time;
} BaseStatistic 

typedef struct FloatStatistics {
    BaseStatistic _base;
    double min_float64;
    double max_float64;
    double first_float64;
    double last_float64;
} FloatStatistic 

May consider patterns like this?
Otherwise, the memory footprint of Statistics will be much higher.

python/tsfile/tsfile_py_cpp.pyx
```diff
@@ -922,3 +927,164 @@ cdef object get_all_timeseries_schema(TsFileReader reader):
...
+        int(m.chunk_meta_count),
+        stat,
+    )
+
+cdef tuple c_device_segments_to_tuple(char** segs, uint32_t n):
+    cdef uint32_t i
+    cdef list out = []
+    for i in range(n):
+        if segs[i] == NULL:
+            out.append("")
```
jt2594838: use none?

jt2594838 (approved):

python/tsfile/tsfile_py_cpp.pyx
hongzhi-gao: fixed

cpp/src/cwrapper/tsfile_cwrapper.h
hongzhi-gao: fixed

PR #769, opened 2026-04-08
Remove timestamps cache to optmize init phase

ycycse:
Main changes:
  - build dataset metadata from `get_timeseries_metadata()` instead of scanning full table data during
  `TsFileSeriesReader` initialization
  - use table-model `query_table_by_row(...)` to support lazy `Timeseries` position access (`series[i]`, small slices)
  - optimize overlapping-shard position reads with chunked k-way merge, avoiding full timestamp materialization for
  small reads
  - remove the incorrect device-level timestamp-axis assumption and keep duplicate timestamp validation on logical-
  series paths
  - fix `query_table_by_row(..., tag_filter=...)` tag-filter handle ownership in Python bindings

  ## C++ note

  This PR also keeps a small C++ reader fix in `tsfile_reader.cc/.h`.

  After `query_table_by_row(...)` was introduced, the same `TsFileReader` could mix row-mode and batch-mode table
  queries. Reusing one `TableQueryExecutor` across those mode switches could leave the reader in the wrong mode and
  break later batch reads.

  To avoid that, `TsFileReader` now recreates `TableQueryExecutor` when the requested table-query mode/configuration
  changes.

### DIFF
--- a/cpp/src/cwrapper/tsfile_cwrapper.cc
+++ b/cpp/src/cwrapper/tsfile_cwrapper.cc
@@ -970,6 +970,79 @@ int fill_timeseries_statistic(storage::Statistic* st,
     return common::E_OK;
 }
 
+int fill_timeline_statistic(storage::ITimeseriesIndex* idx,
+                            TimeseriesStatistic* out) {
+    clear_timeseries_statistic(out);
+    if (idx == nullptr) {
+        return common::E_OK;
+    }
+
+    auto* aligned_idx = dynamic_cast<storage::AlignedTimeseriesIndex*>(idx);
+    if (aligned_idx != nullptr && aligned_idx->time_ts_idx_ != nullptr &&
+        aligned_idx->time_ts_idx_->get_statistic() != nullptr) {
+        auto* st = aligned_idx->time_ts_idx_->get_statistic();
+        TsFileStatisticBase* b = tsfile_statistic_base(out);
+        b->has_statistic = true;
+        b->type = TS_DATATYPE_INVALID;
+        b->row_count = st->get_count();
+        b->start_time = st->start_time_;
+        b->end_time = st->get_end_time();
+        return common::E_OK;
+    }
+
+    if (idx->get_statistic() != nullptr && idx->get_time_chunk_meta_list() == nullptr) {
+        auto* st = idx->get_statistic();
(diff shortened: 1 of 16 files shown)


jt2594838 (approved):

python/tsfile/dataset/timeseries.py
```diff
@@ -101,32 +103,29 @@ def __len__(self) -> int:
...
-            ts = int(timestamps[key])
-            _, values = self._query_time_range(ts, ts)
+            _, values = self._read_by_position(key, 1)
             return float(values[0]) if len(values) > 0 else None
 
         if isinstance(key, slice):
-            requested_ts = timestamps[key]
-            if len(requested_ts) == 0:
+            positions = list(range(*key.indices(length)))
+            if not positions:
```
jt2594838: If the step is 1 or not set, it is unnecessary to list all positions?

python/tsfile/dataset/reader.py
```diff
@@ -138,51 +146,64 @@ def _cache_metadata_table_model(self):
...
+            ]
+            table_groups.sort(key=lambda group: tuple("" if value is None else str(value) for value in group.segments))
+
+            for group in table_groups:
+                stats = self._metadata_device_stats(group)
+                if stats is None:
+                    continue
+                tag_values = self._metadata_tag_values(group, len(tag_columns))
+                if len(tag_values) != len(tag_columns):
+                    continue
```
jt2594838: It is expected that a DeviceId can be a prefix of all tag columns, unnecessary to skip it.

cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -970,6 +970,79 @@ int fill_timeseries_statistic(storage::Statistic* st,
...
+    }
+
+    if (!has_statistic) {
+        return common::E_OK;
+    }
+
+    TsFileStatisticBase* b = tsfile_statistic_base(out);
+    b->has_statistic = true;
+    b->type = TS_DATATYPE_INVALID;
+    b->row_count = row_count;
```
ColinLeeo, line 1042: Should be int64/ timestamp.

ColinLeeo (approved):

PR #772, opened 2026-04-09
Write with parallel

ColinLeeo:
This change introduces thread pool-based write parallelization to the TsFile C++ write path. 

Threading is controlled at compile time via the ENABLE_THREADS CMake option, with parallel_write_enabled_ and write_thread_count_ added to ConfigValue for runtime control. 

The core approach: write_table() first sorts the Tablet by device (via the new Tablet::sort_by_device()) to make same-device rows contiguous, then precomputes page boundaries from page_writer_max_point_num_ and the current time page's existing point count, and finally dispatches the time column and all value columns into the thread pool in parallel — each column writes the same row segments and explicitly seals pages at the precomputed boundaries, preserving time/value page alignment required by the aligned model. 

On the testing side, TsFileWriterTableTest has been converted to a gtest parameterized test, running both Serial        
  (parallel_write_enabled_=false) and Parallel (parallel_write_enabled_=true) suites — all 34 test cases pass under a single build configuration.

### DIFF
--- a/cpp/src/common/global.cc
+++ b/cpp/src/common/global.cc
@@ -24,6 +24,8 @@
 #endif
 #include <stdlib.h>
 
+#include <thread>
+
 #include "utils/injection.h"
 
 namespace common {
@@ -61,8 +63,11 @@ void init_config_value() {
 #else
     g_config_value_.default_compression_type_ = UNCOMPRESSED;
 #endif
-    g_config_value_.parallel_write_enabled_ = true;
-    g_config_value_.write_thread_count_ = 6;
+    unsigned int hw_cores = std::thread::hardware_concurrency();
+    if (hw_cores == 0) hw_cores = 1;  // fallback if detection fails
+    g_config_value_.parallel_write_enabled_ = (hw_cores > 1);
+    g_config_value_.write_thread_count_ =
+        static_cast<int32_t>(std::min(hw_cores, 64u));
     // Enforce aligned page size limits strictly by default.
     g_config_value_.strict_page_size_ = true;
 }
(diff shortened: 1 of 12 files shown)


cpp/src/common/global.cc
```diff
@@ -54,12 +54,15 @@ void init_config_value() {
...
     g_config_value_.double_encoding_type_ = GORILLA;
+    g_config_value_.string_encoding_type_ = PLAIN;
     // Default compression type is LZ4
 #ifdef ENABLE_LZ4
     g_config_value_.default_compression_type_ = LZ4;
 #else
     g_config_value_.default_compression_type_ = UNCOMPRESSED;
 #endif
+    g_config_value_.parallel_write_enabled_ = true;
+    g_config_value_.write_thread_count_ = 6;
```
jt2594838: Is it possible to sense the number of cores?
We may forget to disable parallel-write for end scenarios, and it would be better to auto-disable parallel-write when there is only one core.

cpp/src/common/global.h
```diff
@@ -163,6 +163,20 @@ FORCE_INLINE uint8_t get_global_compression() {
     return static_cast<uint8_t>(g_config_value_.default_compression_type_);
 }
 
+FORCE_INLINE void set_parallel_write_enabled(bool enabled) {
+    g_config_value_.parallel_write_enabled_ = enabled;
+}
+
+FORCE_INLINE bool get_parallel_write_enabled() {
+    return g_config_value_.parallel_write_enabled_;
+}
```
jt2594838, line 173: && g_config_value_.write_thread_count_ != 1?

jt2594838 (commented):
Introduce this feature in ReadMe

jt2594838 (approved):

PR #773, opened 2026-04-10
Optimize and Fix some bugs in TsFileDataFrame

ycycse:
1. Optmize the performance of print(tsdf) and tsdf.show()
2. Optmize loading experience when metadata is huge (eg. millions timeseries)
3. Fix loc return more data than we need
4. Fix temperal way to fix query by row problem. (Real fix should be in tsfile c++ api)
5. Support the situation when tag is None

### DIFF
--- a/python/tsfile/dataset/dataframe.py
+++ b/python/tsfile/dataset/dataframe.py
@@ -764,20 +764,44 @@ class TsFileDataFrame:
                 seen.setdefault(column, True)
         return list(seen.keys())
 
+    @staticmethod
+    def _preview_indices(indices: List[int], max_rows: int) -> Tuple[List[int], bool, int]:
+        total = len(indices)
+        if total <= max_rows:
+            return indices, False, total
+
+        head = max_rows // 2
+        tail = max_rows - head
+        return list(indices[:head]) + list(indices[-tail:]), True, head
+
     def _format_table(self, indices=None, max_rows: int = 20) -> str:
-        series_names = []
-        merged_info = {}
-        for series_ref in self._index.series_refs_ordered:
-            series_name = self._build_series_name(series_ref)
-            series_names.append(series_name)
-            merged_info[series_name] = self._build_series_info(series_ref)
+        if indices is None:
+            indices = list(range(len(self._index.series_refs_ordered)))
+        else:
+            indices = list(indices)
+
+        preview_indices, truncated, split_index = self._preview_indices(indices, max_rows)
+        rows = []
+        for idx in preview_indices:
+            series_ref = self._index.series_refs_ordered[idx]
+            info = self._build_series_info(series_ref)
+            row = {
+                "index": idx,
+                "table": info["table_name"],
+                "field": info["field"],
+                "start_time": info["min_time"],
+                "end_time": info["max_time"],
+                "count": info["count"],
+            }
+            row.update(info["tag_values"])
+            rows.append(row)
 
         return format_dataframe_table(
-            series_names,
-            merged_info,
+            rows,
             self._collect_tag_columns(),
-            indices=indices,
-            max_rows=max_rows,
+            total_count=len(indices),
+            truncated=truncated,
+            split_index=split_index,
         )
 
     def _repr_header(self) -> str:
--- a/python/tsfile/dataset/dataframe.py
+++ b/python/tsfile/dataset/dataframe.py
@@ -29,7 +29,12 @@ import warnings
 import numpy as np
 
 from .formatting import format_dataframe_table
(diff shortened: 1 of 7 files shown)


python/tsfile/dataset/dataframe.py
```diff
@@ -557,14 +576,31 @@ def _load_metadata(self):
         if not self._index.series_refs_ordered:
             raise ValueError("No valid time series found in the provided TsFile files")
 
+    def _show_loading_progress(self, done: int, total: int, total_series: int = None):
+        if not self._show_progress or total <= 0:
+            return
+
+        if total_series is None:
+            sys.stderr.write(f"\rLoading TsFile shards: {done}/{total}")
```
jt2594838, line 584: Is it proper to report progress in stderr？

python/tsfile/dataset/dataframe.py
```diff
@@ -664,11 +705,26 @@ def __len__(self) -> int:
...
+        try:
+            prefix_parts = split_logical_series_path(path_prefix)
+        except ValueError:
+            return []
+
+        matched = []
+        for series_ref in self._index.series_refs_ordered:
+            device_key, table_entry, field_idx = self._get_series_components(series_ref)
+            components = build_logical_series_components(
+                table_entry.table_name,
```
jt2594838, line 720: May use device_key[0] for better readability.
It is confusing at first glance that it suddenly  to [1]

python/tsfile/dataset/metadata.py
```diff
@@ -148,17 +170,38 @@ def split_logical_series_path(series_path: str) -> List[str]:
...
     return _PATH_SEPARATOR.join(_escape_path_component(component) for component in components)
 
 
+def build_logical_series_components(
+    table_name: str,
+    tag_values: Iterable[Any],
+    field_name: str,
+    tag_columns: Iterable[str] = (),
+) -> List[str]:
+    del tag_columns
```
jt2594838: what is this

python/tsfile/dataset/metadata.py
```diff
@@ -187,25 +230,45 @@ def resolve_series_path(catalog: MetadataCatalog, series_path: str) -> Tuple[int
...
+
+    if table_id not in catalog.tables_with_sparse_tag_values:
+        if direct_device_id is None:
+            raise ValueError(f"Series not found: {series_path}")
+        return table_id, direct_device_id, field_idx
+
+    compressed_key = (table_id, tuple(tag_parts))
+    sparse_device_ids = catalog.sparse_device_ids_by_compressed_path.get(compressed_key, [])
+    candidate_ids = []
+    seen_ids = set()
```
jt2594838, line 256: is it enough to use seen_ids only？

python/tsfile/dataset/reader.py
```diff
@@ -47,6 +47,8 @@ def _to_python_scalar(value):
 def _build_exact_tag_filter(tag_values: Dict[str, object]):
     tag_filter = None
     for tag_column, tag_value in tag_values.items():
+        if tag_value is None:
+            continue
```
jt2594838: why should none be skipped？
if the provided tag values are （a， none，c）
then （a，b，c）should not be matched

python/tsfile/dataset/dataframe.py
ycycse: progress output intentionally goes to stderr to keep stdout clean

python/tsfile/dataset/dataframe.py
ycycse: `device_key` is a `(table_name, tag_values)` tuple here, so `device_key[1]` means the tag values. I will expand it to a named local variable to make the code easier to read.

python/tsfile/dataset/metadata.py
ycycse: This is just there to mark the parameter as intentionally unused. I agree the current form is not clear. I will replace it with a better way.

python/tsfile/dataset/metadata.py
ycycse: Yes. `device_id` is the identity we need to deduplicate here. If multiple distinct device ids still remain, we keep that ambiguity and raise an error.

python/tsfile/dataset/reader.py
ycycse: Get. But the current native tag filter does not support `IS NULL` / `IS NOT NULL`, and the row-based path depends on native filtering to preserve lazy `offset/limit` reads. For now I changed this to fail fast with a clear error instead of silently broadening the match. Proper support should be added in the native tag filter layer.

jt2594838 (approved):

PR #793, opened 2026-04-23
feat(tools): support Parquet and Arrow import with schema/auto modes

gengziyand:
## Summary
  - Add Parquet and Arrow format support to the TsFile import tool (alongside existing CSV)
  - Add auto mode (schema-less) import for all three formats
  - Unify schema naming: `id_columns` → `tag_columns`, `csv_columns` → `source_columns` (backward compatible)
  - Add `parquet2tsfile.sh/.bat` and `arrow2tsfile.sh/.bat` scripts
  - Add CLI options: `--format`, `--table_name`, `--time_precision`, `--separator`

  ## Changes
  - New classes: `ParquetSourceReader`, `ArrowSourceReader`, `AutoSchemaInferer`, `ImportExecutor`, `ImportSchema`, `ImportSchemaParser`, `SourceReader`,
  `SourceBatch`, `TabletBuilder`, `TimeConverter`, `ValueConverter`
  - Modified: `TsFileTool.java` (unified CLI entry point), `CsvSourceReader.java` (auto mode support)
  - New dependencies: `parquet-hadoop 1.14.4`, `hadoop-common 3.3.6`, `arrow-vector 15.0.2`
  - 189 automated tests across 11 test classes
  - Updated README.md and README-zh.md

  ## Test plan
  - [x] 189 unit and E2E tests passing (JDK 8)
  - [x] Smoke test on Linux JDK 17 (Docker) - 9 scenarios
  - [x] Smoke test on Windows JDK 8 and JDK 17 - 9 scenarios
  - [x] Data correctness verification - 39 checks (table name, TAG/FIELD, cross-format consistency)
  - [x] Boundary tests - 8 scenarios (tab separator, null values, SKIP/DEFAULT, error handling, fail_dir)

### DIFF
--- a/java/tools/README.md
+++ b/java/tools/README.md
@@ -162,7 +162,7 @@ Omit `--schema` to automatically infer column types and detect the time column.
 **Auto mode rules:**
 - Time column: must be named exactly `time` or `TIME` (case-sensitive, strict match)
 - All other columns become FIELD (no tag inference)
-- CSV type inference uses a 100-row sampling window with promotion chain: `BOOLEAN → INT64 → DOUBLE → STRING`
+- CSV type inference uses a 100-row sampling window. Promotion rules: INT64 and DOUBLE promote to DOUBLE; any other mixed pair (including BOOLEAN with numeric) promotes to STRING.
 - Parquet / Arrow use native schema types directly
 - Default table name: derived from source filename (e.g. `sensor.csv` → table `sensor`)
 - Default null tokens (CSV only): empty cell and `\N`
--- a/java/tools/README.md
+++ b/java/tools/README.md
@@ -44,55 +44,41 @@ mvn clean package -P with-java -DskipTests
 mvn install -P with-java -DskipTests
 ```
 
-## schema 定义
-| Parameter             | Description                      | Required | Default  |
-|----------------|--------------------------|----------|------|
-| table_name     | Table name                       | Yes      |      |
-| time_precision | Time precision (options: ms/us/ns) | No       | ms   |
-| has_header     | Whether it contains a header (options: true/false) | No       | true |
-| separator      | Delimiter (options: , /tab/ ;)    | No        | ,    |
-| null_format    | Null value                       | No        |    |
-| id_columns     | Primary key columns, supports columns not in the CSV as hierarchy     | No        |      |
-| time_column    | Time column                      | Yes        |      |
-| csv_columns    | Corresponding columns in the CSV in order            | Yes        |      |
+## Schema Definition
 
-Explanation:
+### Parameters
 
-The "id_columns" sets values in order and supports using columns that do not exist in the CSV file as levels. 
-For example, if the CSV file has only five columns: "a", "b", "c", "d", and "time",
-id_columns
-a1 default aa
-a
-Among them, a1 is not in the CSV column and is a virtual column with a default value of aa
+| Parameter | Description | Required | Default |
+|-----------|------------|----------|---------|
+| table_name | Table name | Yes | |
+| time_precision | Time precision (ms / us / ns / s) | No | ms |
+| has_header | Whether CSV contains a header (true / false). Ignored for Parquet / Arrow. | No | true |
+| separator | CSV delimiter (, / tab / ;). Ignored for Parquet / Arrow. | No | , |
+| null_format | String value treated as null in CSV. Ignored for Parquet / Arrow (native null). | No | |
+| tag_columns | Tag columns (device identifiers / primary key). Supports virtual columns with DEFAULT value. | No | |
+| time_column | Time column name | Yes | |
+| source_columns | Column definitions mapping to source file columns | Yes | |
 
-The content after csv_columns is the definition of the value column, with the first field in each row being the measurement point name in tsfile and the second field being the type
-When a column in CSV does not need to be written to tsfile, it can be set to SKIP.
+> **Backward compatibility**: `id_columns` and `csv_columns` are still accepted as aliases for `tag_columns` and `source_columns`.
 
-Example:
-csv_columns
-Region TEXT,
-Factory Number TEXT,
-Device Number TEXT,
-SKIP,
-SKIP,
-Time INT64,
-Temperature FLOAT,
-Emission DOUBLE,
+### Column Concepts
 
-Data Example
-CSV file content:
+- **time_column**: Exactly one per table. Written as `time` column with type `TIMESTAMP` in TsFile.
+- **tag_columns**: Device identifiers (composite primary key), 0 or more. Supports virtual columns not present in the source file via `DEFAULT` keyword.
+- **source_columns**: Maps every column in the source file by position (CSV) or by name (Parquet / Arrow). Use `SKIP` to ignore a column.
+- **FIELD** (derived, not configured): All columns in `source_columns` that are not `time_column`, not in `tag_columns`, and not `SKIP`. These are the measurement columns whose values change over time.
 
-### sample data
+### Schema Example
 
 CSV file content:
 ```
 Region,FactoryNumber,DeviceNumber,Model,MaintenanceCycle,Time,Temperature,Emission
-hebei,1001, 1,10,1,1,80.0,1000.0
+hebei,1001,1,10,1,1,80.0,1000.0
 hebei,1001,1,10,1,4,80.0,1000.0
 hebei,1002,7,5,2,1,90.0,1200.0
 ```
-Schema definition
 
+Schema file (`import.schema`):
 ```
 table_name=root.db1
 time_precision=ms
@@ -100,8 +86,7 @@ has_header=true
 separator=,
 null_format=\N
 
-
-id_columns
+tag_columns
 Group DEFAULT Datang
 Region
 FactoryNumber
@@ -109,8 +94,8 @@ DeviceNumber
 
 time_column=Time
 
-csv_columns
-RegionTEXT,
+source_columns
+Region TEXT,
 FactoryNumber TEXT,
 DeviceNumber TEXT,
 SKIP,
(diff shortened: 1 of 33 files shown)


java/tools/src/main/java/org/apache/tsfile/tools/ParquetSourceReader.java
```diff
@@ -0,0 +1,315 @@
...
+    if (parquetReader == null) {
+      parquetReader = ParquetFileReader.open(new LocalInputFile(sourceFile.toPath()));
+      parquetSchema = parquetReader.getFooter().getFileMetaData().getSchema();
+    }
+  }
+
+  private List<String> getSchemaColumnNames() {
+    List<String> names = new ArrayList<>();
+    for (ImportSchema.SourceColumn col : schema.getSourceColumns()) {
+      if (!col.isSkip()) {
```
CritasWang: In schema mode these readers drop skipped source columns from the SourceBatch, but TabletBuilder still indexes values by the original source_columns positions. With a schema like time, unused SKIP, value, getSchemaColumnNames() returns only [time, value], while TabletBuilder maps value to index 2, so batch.getValue(row, 2) throws and the import fails (or later columns shift incorrectly). This affects real Parquet/Arrow schemas using SKIP, and the current Parquet test only checks readBatch(), not the ImportExecutor path. Keep placeholder columns for skipped entries or change the builder mapping to use batch column names rather than schema positions.

CritasWang (changes requested):
need  check skip

CritasWang (approved):

JackieTien97 (dismissed):
## Overall

Good refactoring — the new `SourceReader` / `ImportSchema` / `ImportExecutor` abstraction cleanly separates format-specific reading from the TsFile writing pipeline, and the auto-mode schema inference is a nice usability improvement. The test coverage (189 tests across 11 classes) is solid. Below are the issues I found.

---

### Bug: `time_precision` validation rejects `"s"` but the rest of the code supports it

`ImportSchemaParser.validate()` only allows `ms/us/ns`:

```java
if (!"ms".equals(tp) && !"us".equals(tp) && !"ns".equals(tp)) {
    throw new IllegalArgumentException("time_precision must be ms, us, or ns");
}
```

But `TimeConverter`, the CLI `--time_precision` option, the README, and `ArrowSourceReader.detectTimestampPrecision()` (which returns `"s"` for `SECOND`) all support `"s"`. A schema file with `time_precision=s` will fail at validation.

**Fix:** Add `&& !"s".equals(tp)` to the check, or extract a `VALID_PRECISIONS` set.

---

### Resource leak in `ArrowSourceReader.ensureReaderOpen()`

```java
arrowReader = new ArrowFileReader(new FileInputStream(sourceFile).getChannel(), allocator);
```

The `FileInputStream` is created as an anonymous temporary — only its channel is captured. The stream itself is never closed. Even if `ArrowFileReader.close()` closes the channel, the `FileInputStream` may hold OS-level resources.

**Fix:** Store the `FileInputStream` as a field and close it in `ArrowSourceReader.close()`.

---

### `enforcer.skip=true` suppresses Maven enforcer

```xml
<enforcer.skip>true</enforcer.skip>
```

For an Apache project, silently skipping the enforcer hides dependency convergence issues. The heavy Hadoop dependency tree likely triggers this — please either resolve the convergence conflicts explicitly, or add a comment explaining why it's safe to skip.

---

### Dependency versions should use properties

`parquet-hadoop 1.14.4`, `hadoop-common 3.3.6`, `arrow-vector 15.0.2` are hardcoded in `pom.xml`. These should be managed via `<properties>` (ideally in the parent POM) for consistent version management. This also makes upgrades more visible.

---

### `ImportExecutor.execute()` accepts `failDir` but never uses it

```java
public boolean execute(SourceReader reader, String outputDir, String sourceBaseName, String failDir) {
```

The `failDir` parameter is never read in the method body. Either wire it up (e.g., move failed source files there on error) or remove it to avoid misleading callers.

---

### Documentation mismatch on type promotion chain

The README states:
> CSV type inference uses a 100-row sampling window with promotion chain: `BOOLEAN → INT64 → DOUBLE → STRING`

This implies BOOLEAN promotes to INT64, but the code (`AutoSchemaInferer.promote()`) sends BOOLEAN mixed with any other type directly to STRING:

```java
if (current == InferredType.BOOLEAN || incoming == InferredType.BOOLEAN) {
    return InferredType.STRING;
}
```

The actual behavior is: INT64 ↔ DOUBLE promotes to DOUBLE; any other mixed pair promotes to STRING. The docs should reflect this.

---

### Minor

- `TagColumn.existsInSource()` returns `!hasDefault`, which means any column with a `DEFAULT` value is treated as virtual (not in source). The method name is slightly misleading — consider renaming to `isVirtual()` with inverted semantics, or adding a clarifying comment.
- In `CsvSourceReader.readBatch()`, when the chunk size boundary is hit, the boundary line is both added to `rows` and triggers a return — but `currentSize` is not updated to include this final line, which subtly makes chunks slightly larger than the configured size. Not a functional issue, just an off-by-one on the accounting.

java/tools/src/main/java/org/apache/tsfile/tools/ImportSchemaParser.java
```diff
@@ -0,0 +1,214 @@
...
+      case "TIMESTAMP":
+        return TSDataType.TIMESTAMP;
+      default:
+        throw new IllegalArgumentException("Unknown data type: " + typeStr);
+    }
+  }
+
+  private static void validate(ImportSchema schema) {
+    String tp = schema.getTimePrecision();
+    if (!"ms".equals(tp) && !"us".equals(tp) && !"ns".equals(tp)) {
```
JackieTien97: **Bug:** This validation rejects `"s"` as a valid `time_precision`, but `TimeConverter`, the CLI `--time_precision` option, the README (which lists `ms / us / ns / s`), and `ArrowSourceReader.detectTimestampPrecision()` (returns `"s"` for `SECOND`) all support it.

A schema file with `time_precision=s` will fail here.

**Suggested fix:**
```java
if (!"ms".equals(tp) && !"us".equals(tp) && !"ns".equals(tp) && !"s".equals(tp)) {
    throw new IllegalArgumentException("time_precision must be ms, us, ns, or s");
}
```
Or extract a `VALID_PRECISIONS` set for single-source-of-truth.

java/tools/src/main/java/org/apache/tsfile/tools/ArrowSourceReader.java
```diff
@@ -0,0 +1,298 @@
...
+    if (allocator != null) {
+      allocator.close();
+      allocator = null;
+    }
+  }
+
+  private void ensureReaderOpen() throws IOException {
+    if (arrowReader == null) {
+      allocator = new RootAllocator();
+      arrowReader = new ArrowFileReader(new FileInputStream(sourceFile).getChannel(), allocator);
```
JackieTien97: **Resource leak:** The `FileInputStream` is created as an anonymous temporary — only its channel is captured. The stream itself is never closed. Even if `ArrowFileReader.close()` closes the channel, the `FileInputStream` object may hold OS-level resources.

**Suggested fix:** store it as a field and close it in `close()`:
```java
private FileInputStream fileInputStream;

private void ensureReaderOpen() throws IOException {
    if (arrowReader == null) {
        allocator = new RootAllocator();
        fileInputStream = new FileInputStream(sourceFile);
        arrowReader = new ArrowFileReader(fileInputStream.getChannel(), allocator);
        // ...
    }
}

public void close() {
    // ... close arrowReader first ...
    if (fileInputStream != null) {
        try { fileInputStream.close(); } catch (IOException e) { ... }
    }
    // ... close allocator ...
}
```

java/tools/pom.xml
```diff
@@ -28,6 +28,9 @@
     </parent>
     <artifactId>tools</artifactId>
     <name>TsFile: Java: Tools</name>
+    <properties>
+        <enforcer.skip>true</enforcer.skip>
```
JackieTien97: Silently skipping the Maven enforcer hides dependency convergence issues. The heavy Hadoop/Arrow dependency tree likely triggers this — please either:
1. Resolve the convergence conflicts explicitly (preferred for Apache projects), or
2. Add a comment explaining why it's safe to skip and what specific conflict it works around.

java/tools/pom.xml
```diff
@@ -52,6 +55,87 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-hadoop</artifactId>
+            <version>1.14.4</version>
```
JackieTien97: Hardcoded dependency versions (`parquet-hadoop 1.14.4`, `hadoop-common 3.3.6`, `arrow-vector 15.0.2`) should be managed via `<properties>` — ideally in the parent POM — for consistent version management across modules. This also makes upgrades more visible in diffs.

java/tools/src/main/java/org/apache/tsfile/tools/ImportExecutor.java
```diff
@@ -0,0 +1,134 @@
...
+    this.timeConverter = new TimeConverter(importSchema.getTimePrecision());
+    this.tabletBuilder = new TabletBuilder(importSchema, timeConverter);
+  }
+
+  public boolean execute(SourceReader reader, String outputDir, String sourceBaseName) {
+    return execute(reader, outputDir, sourceBaseName, null);
+  }
+
+  public boolean execute(
+      SourceReader reader, String outputDir, String sourceBaseName, String failDir) {
```
JackieTien97: `failDir` parameter is accepted but never used in the method body. This is misleading to callers who pass a fail directory expecting it to be wired up.

Either use it (e.g., move/copy failed source files there on error) or remove the parameter.

java/tools/README.md
```diff
@@ -119,9 +104,108 @@ Time INT64,
...
+```
+
+### Auto Mode
+
+Omit `--schema` to automatically infer column types and detect the time column.
+
+**Auto mode rules:**
+- Time column: must be named exactly `time` or `TIME` (case-sensitive, strict match)
+- All other columns become FIELD (no tag inference)
+- CSV type inference uses a 100-row sampling window with promotion chain: `BOOLEAN → INT64 → DOUBLE → STRING`
```
JackieTien97: **Doc mismatch with code:** This implies BOOLEAN promotes to INT64, but `AutoSchemaInferer.promote()` sends BOOLEAN mixed with **any** other type directly to STRING:
```java
if (current == InferredType.BOOLEAN || incoming == InferredType.BOOLEAN) {
    return InferredType.STRING;
}
```
The actual behavior: INT64 ↔ DOUBLE promotes to DOUBLE; any other mixed pair (including BOOLEAN + numeric) promotes to STRING. The promotion chain description should reflect this.

java/tools/src/main/java/org/apache/tsfile/tools/ImportSchema.java
```diff
@@ -0,0 +1,247 @@
...
+
+    public boolean hasDefault() {
+      return hasDefault;
+    }
+
+    public String getDefaultValue() {
+      return defaultValue;
+    }
+
+    public boolean existsInSource() {
```
JackieTien97: Nit: the name `existsInSource()` is a bit misleading — it actually returns `!hasDefault`, meaning "is not a virtual column." Consider renaming to `isVirtual()` (with inverted return) or at minimum adding a one-liner clarifying the semantics.

java/tools/src/main/java/org/apache/tsfile/tools/CsvSourceReader.java
```diff
@@ -0,0 +1,290 @@
...
+        bufferedSampleRows = null;
+      }
+
+      if (!exhausted) {
+        String line;
+        while ((line = reader.readLine()) != null) {
+          byte[] lineBytes = line.getBytes(StandardCharsets.UTF_8);
+          long lineSize = lineBytes.length;
+
+          if (currentSize > 0 && currentSize + lineSize > chunkSizeBytes) {
```
JackieTien97, line 183: Minor: when the chunk boundary is hit, the boundary line is added to `rows` (line 184) and triggers a return — but `currentSize` is never updated to include this final line. This means chunks can be slightly larger than `chunkSizeBytes`. Not a functional issue (just an off-by-one on the accounting), but worth noting.

JackieTien97 (changes requested):
Overall the new `SourceReader` / `ImportSchema` / `ImportExecutor` abstraction is a clean design that separates format-specific reading from TsFile writing. Auto-mode schema inference is a nice usability win. Test coverage (189 tests across 11 classes) is solid. See inline comments for specific issues.

java/tools/src/main/java/org/apache/tsfile/tools/ImportSchemaParser.java
gengziyand: Fixed. time_precision validation now also accepts "s", and the error message was updated accordingly.

java/tools/src/main/java/org/apache/tsfile/tools/ArrowSourceReader.java
gengziyand: Done. I now keep the FileInputStream as a field in ArrowSourceReader and close it explicitly in close() instead of creating it anonymously from getChannel().

java/tools/pom.xml
gengziyand: Done. I removed enforcer.skip, reproduced the convergence failures locally, and resolved them by managing the conflicting transitive dependency versions in the parent dependencyManagement.

java/tools/pom.xml
gengziyand: Updated. I moved the tools dependency versions out of java/tools/pom.xml and into the parent java/pom.xml dependencyManagement section for centralized version management.

java/tools/src/main/java/org/apache/tsfile/tools/ImportExecutor.java
gengziyand: Done. I removed the unused failDir parameter by collapsing the 4-arg execute(...) into the real 3-arg method and updated the call sites in TsFileTool accordingly.

java/tools/README.md
gengziyand: Fixed. I updated both README.md and README-zh.md so the documented CSV type-promotion behavior now matches the actual implementation.

java/tools/src/main/java/org/apache/tsfile/tools/ImportSchema.java
gengziyand: Updated as suggested. I renamed existsInSource() to isVirtual(), inverted the return semantics, and adjusted all call sites/tests to match the new meaning.

java/tools/src/main/java/org/apache/tsfile/tools/CsvSourceReader.java
gengziyand: Fixed. I added the missing currentSize update before returning the batch at the chunk boundary.

JackieTien97 (approved):

PR #796, opened 2026-04-30
Fix TS_2DIFF float/double page compatibility

hongzhi-gao:
## Summary

- Fix C++ TS_2DIFF float/double page layout to match Java overflow-page encoding.
- Fix Java table-read time decoder selection to use actual time chunk encoding (instead of global default), preventing mis-decoding on PLAIN time chunks.
- Harden googletest zip handling in CMake: reject/remove empty or invalid zip files and only accept valid ZIP magic before enabling tests.

## Encoding Layout

**Legacy raw block (old C++ path):**  
[ TS_2DIFF inner block ]

**Java-compatible overflow page (fixed C++ path):**  
[ outer_magic(varint=2147483646) ]  
[ count(varint) ]  
[ bitmap_under ]  
[ bitmap_over ]  
[ inner_max_point(varint=2) ]  
[ TS_2DIFF inner block payload ]

## Test Input

Float values: `3.123456768E20f`, `NaN`  
Expected encoded hex (Java baseline):

`FE FF FF FF 07 02 00 03 02 00 00 00 01 00 00 00 00 1E 38 8A AA 61 87 75 56`

## Verification

- C++ encoded output matched the Java baseline byte-for-byte.
- Java table query can read both C++-generated and Java-generated files


cpp/src/encoding/ts2diff_encoder.h
```diff
@@ -402,5 +484,168 @@ FORCE_INLINE int DoubleTS2DIFFEncoder::encode(double value,
...
+            } else if (f == -1) {
+                has_value_overflow = true;
+                value_overflow_bitmap[static_cast<size_t>(i / 8)] |=
+                    static_cast<uint8_t>(1u << (i % 8));
+            }
+        }
+        constexpr uint32_t kJavaOverflowMagic =
+            2147483647u;  // Integer.MAX_VALUE
+        constexpr uint32_t kJavaValueOverflowMagic =
+            2147483646u;  // Integer.MAX_VALUE - 1
```
jt2594838: kJavaOverflowMagic -> FLAG_SCALED_VALUE_OVERFLOW
kJavaValueOverflowMagic  -> FLAG_ORIGINAL_VALUE_OVERFLOW

cpp/src/encoding/ts2diff_encoder.h
```diff
@@ -402,5 +484,168 @@ FORCE_INLINE int DoubleTS2DIFFEncoder::encode(double value,
...
+    }
+    flush_remaining(inner);
+    reset();
+
+    const bool overflow = has_overflow();
+    if (overflow) {
+        std::vector<uint8_t> underflow_bitmap(
+            static_cast<size_t>(num_values / 8 + 1), 0);
+        std::vector<uint8_t> value_overflow_bitmap(
+            static_cast<size_t>(num_values / 8 + 1), 0);
```
jt2594838, line 607: May use overflow/underflow consistently.

jt2594838 (approved):

cpp/src/encoding/ts2diff_encoder.h
hongzhi-gao: Renamed as suggested

cpp/src/encoding/ts2diff_encoder.h
hongzhi-gao: Updated

jt2594838 (approved):

PR #815, opened 2026-05-09
Optimize get_timeseries_metadata() to eliminate N+1 I/O (~94x speedup)

JackieTien97:
## Summary

Optimize TsFile C++ read path in two areas:

### 1. Optimize `get_timeseries_metadata()` — eliminate N+1 I/O (~94x speedup for init)

**Root Causes:**
1. **N+1 I/O**: `get_all_device_ids()` traverses the entire device index tree but discards offset information. Then for each device, `load_device_index_entry()` re-searches the tree from the root — O(N×D) redundant disk reads for N devices and tree depth D.
2. **Duplicate read**: `get_device_timeseries_meta_without_chunk_meta()` reads the measurement index node once to check alignment, then `load_all_measurement_index_entry()` reads the exact same byte range again.
3. **Redundant PageArena::init()**: Called per-device in a loop despite already being initialized in the constructor.

**Changes:**
- Add `TsFileIOReader::get_device_timeseries_meta_by_offset()` — accepts pre-resolved offsets (skips `load_device_index_entry`), reuses deserialized top node for `get_all_leaf()` (eliminates duplicate read)
- Add `TsFileReader::get_all_device_entries()` — single tree traversal collecting device IDs with their `(start_offset, end_offset)`
- Rewrite `get_timeseries_metadata()` to combine the above
- Remove redundant `PageArena::init()` call from `get_timeseries_metadata_impl()`

**Benchmark** (`ecg_dataset/part_0.tsfile`, 634 MB, 53040 devices):

| Path | Avg Time | Speedup |
|------|----------|---------|
| **Before** (`get_all_device_ids` + `get_timeseries_metadata(ids)`) | ~28.8 s | — |
| **After** (`get_timeseries_metadata()` optimized) | ~0.30 s | **~94x** |

### 2. Optimize `queryByRow` with exact tag filter — B-tree direct lookup (~230x speedup per query)

**Root Cause:**
When `queryByRow` is called with a tag filter (e.g. exact match on one device), `DeviceMetaIterator` still traverses the entire MetaIndexNode tree: reading all 208 internal nodes from disk and scanning all 53040 leaf device entries to check the filter. This takes ~117ms per query even though only 1 device matches.

**Changes:**
- `DeviceMetaIterator::try_setup_direct_lookup()` — detects single `TagEq` filter at construction time and constructs the target device ID
- `DeviceMetaIterator::load_results_direct()` — uses `load_device_index_entry()` for O(log N) B-tree binary search instead of full tree traversal
- Move `load_device_index_entry()` from private to public in `TsFileIOReader` for reuse
- Add `bench_read.cpp` for standalone C++ read path benchmarking

**Benchmark** (`ecg_dataset/part_0.tsfile`, 53040 devices, offset=1000, limit=3584):

| Component | Before | After | Speedup |
|-----------|--------|-------|---------|
| `DeviceTaskIterator::next()` | 117.4 ms | 0.5 ms | **~230x** |
| `first next()` (full lazy init) | 128 ms | 1.8 ms | **~71x** |
| Total `queryByRow` | 128 ms | 1.9 ms | **~67x** |

**End-to-end Python benchmark** (training sample read, ecg_dataset):

| Metric | Before | After | Speedup |
|--------|--------|-------|---------|
| Average latency | 236.8 ms | 5.74 ms | **41x** |
| Throughput | 4.32 windows/sec | 174.1 windows/sec | **40x** |

> Note: The single-TagEq direct lookup covers the common training data access pattern. Multi-tag AND filters fall back to the existing full-scan path.

## Test plan

- [x] All 522 existing tests pass
- [x] Verified result correctness: both paths return identical device count (53040)
- [x] C++ bench_read validates queryByRow returns correct row count (3584)
- [x] Python end-to-end benchmark validates data correctness with training pipeline


cpp/src/reader/tsfile_reader.cc
```diff
@@ -25,6 +25,69 @@
...
+                (i + 1 < index_node->children_.size())
+                    ? index_node->children_[i + 1]->get_offset()
+                    : index_node->end_offset_;
+            entries.push_back(entry);
+        }
+    } else {
+        for (size_t idx = 0; idx < index_node->children_.size(); idx++) {
+            auto meta_index_entry = index_node->children_[idx];
+            int start_offset = meta_index_entry->get_offset();
+            int end_offset = index_node->end_offset_;
```
JackieTien97: **Bug: `int64_t` truncated to `int`**

`get_offset()` returns `int64_t` and `end_offset_` is `int64_t`. Assigning them to `int` (32-bit) silently truncates for files >2 GB — the read offset passed to `read_file->read()` will be wrong.

Both should be `int64_t`:
```cpp
int64_t start_offset = meta_index_entry->get_offset();
int64_t end_offset = index_node->end_offset_;
```

cpp/src/file/tsfile_io_reader.cc
```diff
@@ -155,6 +155,70 @@ int TsFileIOReader::get_device_timeseries_meta_without_chunk_meta(
...
+    int ret = E_OK;
+    load_tsfile_meta_if_necessary();
+
+    std::vector<std::pair<std::shared_ptr<IMetaIndexEntry>, int64_t>>
+        meta_index_entry_list;
+    bool is_aligned = false;
+    TimeseriesIndex* time_timeseries_index = nullptr;
+
+    ASSERT(start_offset < end_offset);
+    const int32_t read_size = end_offset - start_offset;
```
JackieTien97, line 170: Nit: `end_offset - start_offset` is `int64_t` arithmetic, implicitly narrowed to `int32_t`. For a single measurement index node the size is unlikely to exceed 2 GB so this is fine in practice, and it matches the `int32_t buf_size` parameter of `ReadFile::read()`. Just noting for awareness.

JackieTien97 (commented):
## Code Review

Great optimization — the 94x speedup from eliminating the N+1 I/O pattern is impressive. The approach of collecting device entries with offsets in a single traversal and then reusing the deserialized top node is well-designed.

One **critical bug** found: integer truncation of 64-bit file offsets in `get_all_device_entries`. See inline comments.

ColinLeeo:
LGTM

ColinLeeo (approved):

jt2594838 (approved):

PR #817, opened 2026-05-18
support msvc 2017.

ColinLeeo:


cpp/src/file/read_file.cc
```diff
@@ -21,10 +21,15 @@
 
 #include <fcntl.h>
 #include <sys/stat.h>
+#ifdef _WIN32
+#include <io.h>
```
hongzhi-gao, line 25: Duplicate #include <io.h> on _WIN32 — please remove one.

cpp/patches/antlr4-4.9.3-msvc-compat.patch
jt2594838, line 1: What is this?

cpp/CMakeLists.txt
```diff
@@ -178,7 +231,15 @@ set(LIBRARY_INCLUDE_DIR ${PROJECT_BINARY_DIR}/include CACHE STRING "TsFile inclu
 set(THIRD_PARTY_INCLUDE ${PROJECT_BINARY_DIR}/third_party)
 
 set(SAVED_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-set(CMAKE_CXX_FLAGS "$ENV{CXXFLAGS} -Wall -std=c++11")
+if (MSVC)
+    # MSVC does not provide a /std:c++11 flag; C++11 is its implicit baseline.
+    # The lowest explicitly settable standard is /std:c++14. Without this flag,
+    # the default varies by VS version (VS2017+ defaults to C++14 mode with some
+    # C++17 extensions), so we pin it explicitly for reproducibility.
+    set(CMAKE_CXX_FLAGS "$ENV{CXXFLAGS} /W3 /utf-8 /EHsc /bigobj /Zc:__cplusplus /std:c++14")
```
jt2594838, line 239: If we specify C++14, will the compilation on old MSVC (2013, for example) fail?

cpp/patches/antlr4-4.9.3-msvc-compat.patch
ColinLeeo: ANTLR4's C++ runtime doesn't compile on MSVC 2017 (upstream supports MSVC 2019+), so minimal source fixes are required to build it in our environment. The changes are kept as a standalone patch — even though the tree is git-tracked — to clearly isolate our modifications to third-party code, making them easy to review, re-apply on version bumps, and drop once we move off MSVC 2017.

cpp/CMakeLists.txt
ColinLeeo: Our code targets C++11; /std:c++14 is used on MSVC 2017 only because MSVC 2017 has no /std:c++11 option (C++14 is its minimum). For older MSVC (2013/2015), the flag should just be omitted via if(MSVC_VERSION GREATER_EQUAL 1910) — though ANTLR4 itself still won't build there.

jt2594838 (approved):

PR #818, opened 2026-05-18
Add AGENTS.md symlinks

HTHou:
## Summary

- Add `AGENTS.md` symlinks wherever `CLAUDE.md` exists.
- Point each symlink at the local `CLAUDE.md` so agent instructions stay in one place.

## Validation

- Verified the symlink targets with `ls -la`.
- Reviewed the staged diff before committing.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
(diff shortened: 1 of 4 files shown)


hongzhi-gao:
good job!

PR #819, opened 2026-05-18
Fix import edge cases for DATE/TIMESTAMP, INT96, CSV quotes and format filtering

gengziyand:
### Summary

  This PR fixes several import edge cases in `java/tools`:

  - Support `DATE` and `TIMESTAMP` value conversion before writing Tablets, and pass import time precision from
  `TabletBuilder` to `ValueConverter`.
  - Decode legacy Parquet `INT96` timestamps from nanos-of-day plus Julian day, and mark them as nanosecond precision in
  auto schema.
  - Resolve timezone offsets from the parsed local datetime and configured `ZoneId`, instead of using the JVM's current
  DST offset.
  - Make `--format` filtering in directory mode respect file extensions, so unrelated files are skipped instead of parsed
  as data.
  - Extract Arrow `DATE` and `TIMESTAMP` vectors through their actual vector classes.
  - Parse quoted CSV fields with embedded separators and escaped quotes.

  ### Tests

  ```bash
  mvn '-Dspotless.apply.skip=true' test
  '-Dtest=ValueConverterTest,TabletBuilderTest,ParquetSourceReaderTest,DateTimeUtilsTest,TsFileToolFormatFilterTest,CsvSou
  rceReaderTest,ArrowSourceReaderTest,TsFileToolEndToEndTest'
  mvn '-Dspotless.apply.skip=true' test

### DIFF
--- a/java/tools/README-zh.md
+++ b/java/tools/README-zh.md
@@ -66,10 +66,15 @@ mvn install -P with-java -DskipTests
 
 - **time_column**：每个表有且仅有一个时间列，写入 TsFile 后列名固定为 `time`，类型为 `TIMESTAMP`。
 - **tag_columns**：设备标识列（联合主键），可以为 0 到多个。支持通过 `DEFAULT` 关键字定义不在源文件中的虚拟列。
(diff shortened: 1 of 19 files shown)


CritasWang (approved):
LGTM

PR #820, opened 2026-05-18
feat: i18n support for log and exception messages

JackieTien97:
## Summary

- Adds runtime `ResourceBundle`-based i18n for all log and exception messages in `java/common`, `java/tsfile`, and `java/tools`
- A single published JAR supports both English (default) and Simplified Chinese; locale is selected at JVM startup via `-Dtsfile.locale=zh` or `Locale.getDefault()`
- 521 message keys, ~900+ call sites migrated across ~110 files
- Includes complete Simplified Chinese translation
- Drift-prevention unit tests enforce key-set alignment between `messages.properties` and `messages_zh.properties` at build time

## Why runtime (vs. compile-time profile)

IoTDB's recent i18n PR ([apache/iotdb#17613](https://github.com/apache/iotdb/pull/17613)) uses a Maven profile to inline locale-specific strings into the JAR at compile time. TsFile cannot adopt that same model without either publishing two artifacts per release or forking the version, since downstreams (IoTDB and others) consume tsfile as a versioned dependency. A runtime ResourceBundle approach keeps a single published artifact while letting downstream IoTDB switch locale via a JVM startup option.

Locale resolution order at JVM startup (once at class load):
1. `-Dtsfile.locale=<tag>` system property (e.g. `zh`, `zh-CN`)
2. `Locale.getDefault()`
3. English fallback via `ResourceBundle`'s root-bundle mechanism

## Architecture

- `org.apache.tsfile.i18n.Messages` (in `java/common`) provides `get(key)` for SLF4J patterns (`{}` placeholders) and `format(key, args)` for exception patterns (`%1$s` positional placeholders).
- A custom `ResourceBundle.Control` loads `.properties` files as UTF-8 (Java 8 default is ISO-8859-1) so Chinese characters can be written directly without `\uXXXX` escapes.
- Resources live at `java/common/src/main/resources/org/apache/tsfile/i18n/messages.properties` (English, root bundle) and `messages_zh.properties` (Chinese).
- `MessagesTest` enforces key-set alignment + all values non-empty in both locales.
- Technical terms (`TsFile`, `Chunk`, `Page`, `Schema`, class/method names, type names, codec names like `RLE`/`GORILLA`) are kept in English in Chinese translations, matching IoTDB's convention.

## IoTDB integration

IoTDB's Chinese build (the `with-zh-locale` profile in apache/iotdb#17613) should add `-Dtsfile.locale=zh` to its startup scripts (e.g. `start-datanode.sh`, `start-confignode.sh`) to also receive Chinese messages from tsfile. No coordinated release is required — IoTDB can adopt this whenever convenient.

## Migration scope

Migrated all literal-string log and exception messages in:

| Module | Keys added |
|--------|------------|
| `common` | 16 |
| `tsfile/write` | 70 |
| `tsfile/read/common` | 51 |
| `tsfile/read` (rest) | 84 |
| `tsfile/encoding` | 58 |
| `tsfile/file` | 17 |
| `tsfile/external` | 65 |
| `tsfile/encrypt` | 13 |
| `tsfile/compress` | 7 |
| `tsfile/fileSystem` | 41 |
| `tsfile/utils` | 37 |
| `tools` | 61 |
| **Total** | **521** (incl. 1 seed) |

Each migration commit is its own logical scope so reviewers can review per-area.

Sites correctly skipped (not migrated):
- Variable/data-only constructor args: `throw new X(deviceId.toString())`, `throw new X(String.valueOf(type))`, `throw new X(getClass().getName())`
- Exception wrapping with no message: `throw new X(e)`
- No-arg throws: `throw new X()`
- Utility methods that pass through caller-supplied format strings (e.g. `Validate.isTrue(boolean, String, Object...)`)

## Test plan

- [x] All existing tests pass under default (English) locale: `./mvnw test -P with-java`
- [x] `MessagesTest` drift-prevention tests pass — verifies en/zh key set alignment + all values non-empty
- [x] `MessagesTest` passes under `-Duser.language=zh -Duser.country=CN` (simulated Chinese locale)
- [x] Spot tests under `-Dtsfile.locale=zh` show Chinese output without test failures
- [x] `./mvnw spotless:check -P with-java` clean
- [x] Key count verified: 521 keys in each `.properties` file

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/encoding/encoder/TSEncodingBuilder.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/encoding/encoder/TSEncodingBuilder.java
@@ -24,6 +24,7 @@ import org.apache.tsfile.common.conf.TSFileDescriptor;
 import org.apache.tsfile.common.constant.JsonFormatConstant;
 import org.apache.tsfile.enums.TSDataType;
 import org.apache.tsfile.file.metadata.enums.TSEncoding;
+import org.apache.tsfile.i18n.Messages;
 import org.apache.tsfile.write.UnSupportedDataTypeException;
 
 import org.slf4j.Logger;
@@ -44,7 +45,7 @@ public abstract class TSEncodingBuilder {
 
   private static final Logger logger = LoggerFactory.getLogger(TSEncodingBuilder.class);
   protected final TSFileConfig conf;
-  private static final String ERROR_MSG = "%s doesn't support data type: %s";
+  private static final String ERROR_MSG_KEY = "error.encoding.ts_encoding_builder_unsupported_type";
 
   protected TSEncodingBuilder() {
     this.conf = TSFileDescriptor.getInstance().getConfig();
@@ -81,7 +82,8 @@ public abstract class TSEncodingBuilder {
       case CAMEL:
         return new Camel();
       default:
-        throw new UnsupportedOperationException("Unsupported encoding: " + type);
+        throw new UnsupportedOperationException(
+            Messages.format("error.encoding.ts_encoding_builder_unsupported", type));
     }
   }
 
@@ -128,8 +130,7 @@ public abstract class TSEncodingBuilder {
         if (maxStringLength < 0) {
           maxStringLength = TSFileDescriptor.getInstance().getConfig().getMaxStringLength();
           logger.warn(
-              "cannot set max string length to negative value, replaced with default value:{}",
-              maxStringLength);
+              Messages.get("log.encoding.ts_encoding_max_string_length_negative"), maxStringLength);
         }
       }
     }
@@ -154,7 +155,8 @@ public abstract class TSEncodingBuilder {
         case DOUBLE:
           return new FloatEncoder(TSEncoding.RLE, type, maxPointNumber);
         default:
-          throw new UnSupportedDataTypeException(String.format(ERROR_MSG, TSEncoding.RLE, type));
+          throw new UnSupportedDataTypeException(
+              Messages.format(ERROR_MSG_KEY, TSEncoding.RLE, type));
       }
     }
 
@@ -172,15 +174,13 @@ public abstract class TSEncodingBuilder {
           this.maxPointNumber = Integer.parseInt(props.get(Encoder.MAX_POINT_NUMBER));
         } catch (NumberFormatException e) {
           logger.warn(
-              "The format of max point number {} is not correct."
-                  + " Using default float precision.",
+              Messages.get("log.encoding.ts_encoding_max_point_number_format"),
               props.get(Encoder.MAX_POINT_NUMBER));
         }
         if (maxPointNumber < 0) {
           maxPointNumber = TSFileDescriptor.getInstance().getConfig().getFloatPrecision();
           logger.warn(
-              "cannot set max point number to negative value, replaced with default value:{}",
-              maxPointNumber);
+              Messages.get("log.encoding.ts_encoding_max_point_number_negative"), maxPointNumber);
         }
       }
     }
@@ -210,7 +210,7 @@ public abstract class TSEncodingBuilder {
           return new FloatEncoder(TSEncoding.TS_2DIFF, type, maxPointNumber);
         default:
           throw new UnSupportedDataTypeException(
-              String.format(ERROR_MSG, TSEncoding.TS_2DIFF, type));
+              Messages.format(ERROR_MSG_KEY, TSEncoding.TS_2DIFF, type));
       }
     }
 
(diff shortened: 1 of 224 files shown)


java/tsfile/src/main/java/org/apache/tsfile/read/TsFileSequenceReader.java
```diff
@@ -2857,7 +2853,7 @@ public List<AbstractAlignedChunkMetadata> getAlignedChunkMetadata(
     Pair<IMetadataIndexEntry, Long> metadataIndexPair =
         getMetadataAndEndOffsetOfDeviceNode(deviceMetadataIndexNode, device, true);
     if (metadataIndexPair == null) {
-      throw new IOException("Device {" + device + "} is not in tsFileMetaData");
+      throw new IOException(Messages.format("error.read.device_not_in_metadata", device));
```
JackieTien97, line 2856: Breaks an existing test under non-English locale.

`TsFileReaderTest.testGetAlignedChunkMetadata` asserts the literal English text returned by this throw:

https://github.com/apache/tsfile/blob/03c3aa6ad5436183c13a0174deeafebad0d89715/java/tsfile/src/test/java/org/apache/tsfile/read/TsFileReaderTest.java#L528-L530

Under `-Dtsfile.locale=zh` (or a Chinese-default JVM), `e.getMessage()` becomes `设备 {d3} 不在 tsFileMetaData 中` from [`error.read.device_not_in_metadata`](https://github.com/apache/tsfile/blob/7c880c3826cb8df4ea74a7d44830e4e51e8fcea0/java/common/src/main/resources/org/apache/tsfile/i18n/messages_zh.properties) in `messages_zh.properties`, and the `assertEquals` fails.

java/tsfile/src/main/java/org/apache/tsfile/read/TsFileSequenceReader.java
```diff
@@ -2886,7 +2882,7 @@ public List<AbstractAlignedChunkMetadata> getAlignedChunkMetadataByMetadataIndex
       throws IOException {
     TimeseriesMetadata firstTimeseriesMetadata = getTimeColumnMetadata(metadataIndexNode, null);
     if (firstTimeseriesMetadata == null) {
-      throw new IOException("Timeseries of device {" + device + "} are not aligned");
+      throw new IOException(Messages.format("error.read.timeseries_not_aligned", device));
```
JackieTien97, line 2885: Same locale-sensitivity issue as the comment above.

`TsFileReaderTest.testGetAlignedChunkMetadata` asserts the literal English text returned by this throw:

https://github.com/apache/tsfile/blob/03c3aa6ad5436183c13a0174deeafebad0d89715/java/tsfile/src/test/java/org/apache/tsfile/read/TsFileReaderTest.java#L536-L538

Under `zh` locale, the message becomes `设备 {d2} 的时间序列不是对齐的` from `error.read.timeseries_not_aligned`, and the `assertEquals` fails.

java/tsfile/src/main/java/org/apache/tsfile/encoding/encoder/TSEncodingBuilder.java
```diff
@@ -81,7 +82,8 @@ public static TSEncodingBuilder getEncodingBuilder(TSEncoding type) {
       case CAMEL:
         return new Camel();
       default:
-        throw new UnsupportedOperationException("Unsupported encoding: " + type);
+        throw new UnsupportedOperationException(
+            Messages.format("error.encoding.ts_encoding_builder_unsupported", type));
```
JackieTien97, line 86: Breaks `TSEncodingBuilderTest.testTSEncodingBuilder` under non-English locale.

The test asserts the literal English text that this throw used to produce:

https://github.com/apache/tsfile/blob/39e55a358291ca756f3ee5d8f85c144f3139ccec/java/tsfile/src/test/java/org/apache/tsfile/encoding/encoder/TSEncodingBuilderTest.java#L55-L58

Under `zh` locale, `e.getMessage()` becomes `不支持的 encoding: %1$s` (with `%1$s` substituted) from `error.encoding.ts_encoding_builder_unsupported`, and the `assertEquals` fails. (Same fix as for the `TsFileSequenceReader` comments.)

JackieTien97 (commented):
### Code review

Found 1 issue (3 affected locations): two existing tests assert exact English exception message text for messages that this PR migrates to i18n, so they break under a non-English JVM locale. Each affected source-line is annotated inline below; the unmodified test files are linked from each comment.

Fix options: pin `System.setProperty("tsfile.locale", "en")` in those tests (same pattern `MessagesTest.java` already uses), or rewrite the assertions to be locale-insensitive.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

HTHou (approved):

PR #749, opened 2026-03-22
Write Opt.

ColinLeeo:

PR #823, opened 2026-05-26
C++: optimize batch read/write paths and aligned table null handling

ColinLeeo:
## Summary

This PR optimizes the C++ TsFile read/write paths for batch and columnar workloads, and fixes several aligned table null-handling issues uncovered while validating the optimized path.

It consolidates the batch decode/write work from the long-lived optimization branch into a reviewable change for `develop`.

Supersedes #749, #754, and #774.

## Main Changes

- Add batch decode APIs to the C++ `Decoder` hierarchy and implement batch paths for PLAIN, TS2DIFF, and Gorilla.
- Add TsBlock-level batch read paths in `ChunkReader` and `AlignedChunkReader`, including shared timestamp decoding for aligned multi-value reads.
- Add columnar tablet write helpers and batch page writes for fixed-length and string columns.
- Add optional parallel page decode/write infrastructure behind the existing C++ configuration.
- Add SIMD-oriented fast paths for TS2DIFF, PLAIN byte swapping, and selected statistics updates.
- Improve `ByteStream`, compressor, and page/chunk writer internals used by the optimized paths.
- Simplify the C wrapper surface by removing unused metadata/tag-filter export APIs.

## Correctness Fixes

- Fix null TAG device boundary detection so null string offsets are not read as valid values.
- Fix null string FIELD writes so null rows do not read uninitialized string offsets.
- Preserve all-null aligned value pages by counting page rows separately from non-null statistic counts.
- Treat missing aligned measurements as sparse columns instead of failing the whole device read.
- Add a time-only aligned read path for devices whose queried FIELD columns are all null/missing.
- Fix repeated logical devices inside one tablet by aggregating their segments before writing aligned chunks.
- Fix `ValuePageWriter::reset()` so row count and null bitmap state are reset together.
- Remove unused temporary TS2DIFF prototype headers that were not part of the production implementation.

## Compatibility Notes

- Existing C++ APIs remain additive for the optimized read/write paths.
- The public C wrapper removes unused metadata export/tag-filter handles. Downstream wrappers should sanity-check that they do not depend on those removed symbols.
- `cpp/third_party/` is left at `develop` state so existing platform compatibility fixes are preserved.

## Verification

- `cmake --build cpp/target/build --target TsFile_Test -j1`
- `ctest --test-dir cpp/target/build/test --output-on-failure -R '^TsFileTableReaderTest\.TestNullInTable4$'`
- `ctest --test-dir cpp/target/build/test --output-on-failure -j4`
- `cd cpp && mvn spotless:check`
- `cd cpp && mvn apache-rat:check`

Current full C++ test result: `496/496` tests pass.


cpp/src/common/allocator/byte_stream.h
```diff
@@ -339,30 +368,46 @@ class ByteStream {
     // never used TODO
     void shallow_clone_from(ByteStream& other) {
         this->page_size_ = other.page_size_;
+        this->page_mask_ = other.page_mask_;
         this->mid_ = other.mid_;
         this->head_.store(other.head_.load());
         this->tail_.store(other.tail_.load());
         this->total_size_.store(other.total_size_.load());
     }
```
ColinLeeo: need delete.

cpp/src/common/container/blocking_queue.cc
```diff
@@ -0,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
```
ColinLeeo, line 3: remove file

cpp/src/common/mutex/mutex.h
```diff
@@ -26,9 +26,6 @@
 
 namespace common {
```
ColinLeeo: remove

cpp/src/common/tablet.cc
```diff
@@ -98,14 +100,8 @@ int Tablet::init() {
             case BLOB:
             case TEXT:
             case STRING: {
-                auto* sc = static_cast<StringColumn*>(common::mem_alloc(
-                    sizeof(StringColumn), common::MOD_TABLET));
-                if (sc == nullptr) return E_OOM;
```
ColinLeeo, line 103: need revert

cpp/src/common/tablet.cc
```diff
@@ -502,75 +577,52 @@ void Tablet::reset_string_columns() {
...
     const uint32_t nwords = (row_count + 63) / 64;
     std::vector<uint64_t> boundary(nwords, 0);
 
-    uint32_t boundary_count = 0;
-    const uint32_t max_boundaries = row_count - 1;
-    for (auto it = id_column_indexes_.rbegin(); it != id_column_indexes_.rend();
-         ++it) {
-        const StringColumn& sc = *value_matrix_[*it].string_col;
-        const int32_t* off = sc.offsets;
+    for (auto col_idx : id_column_indexes_) {
```
ColinLeeo: toreview

cpp/src/common/thread_pool.h
```diff
@@ -98,7 +113,23 @@ class ThreadPool {
...
+            // wrapper installed by std::packaged_task — that path never
+            // reaches here, so this catch only fires for fire-and-forget
+            // tasks where the alternative is termination.
+            try {
+                task();
+            } catch (...) {
+                // Intentionally suppressed; see comment above.
+            }
             {
                 std::lock_guard<std::mutex> lk(mu_);
```
ColinLeeo, line 141: Not good here

cpp/src/encoding/int64_sprintz_decoder.h
```diff
@@ -124,9 +124,8 @@ class Int64SprintzDecoder : public SprintzDecoder {
             decode_size_ = bit_width_ & ~(1 << 7);
             Int64RleDecoder decoder;
             for (int i = 0; i < decode_size_; ++i) {
-                if (RET_FAIL(decoder.read_int(current_buffer_[i], input))) {
-                    return ret;
-                }
+                int ret = decoder.read_int(current_buffer_[i], input);
```
ColinLeeo: need revert

cpp/src/encoding/plain_decoder.h
```diff
@@ -62,6 +99,186 @@ class PlainDecoder : public Decoder {
...
+    // ── Batch overrides ──────────────────────────────────────────────────────
+    //
+    // INT32: PLAIN encoding uses varint (variable stride).  Override to avoid
+    // virtual dispatch per element; actual decode is still per-value.
+    int read_batch_int32(int32_t* out, int capacity, int& actual,
+                         common::ByteStream& in) override {
+        actual = 0;
+        while (actual < capacity && in.has_remaining()) {
+            int ret = common::SerializationUtil::read_var_int(out[actual], in);
+            if (ret != common::E_OK) return ret;
```
ColinLeeo, line 113: Simplify the code using macros.

cpp/src/common/allocator/byte_stream.h
```diff
@@ -231,6 +232,23 @@ FORCE_INLINE double bytes_to_double(uint8_t bytes[8]) {
...
+    if (n <= 1) return 1;
+    if (n > 0x80000000u) return 0x80000000u;
+    uint32_t v = n - 1;
+    v |= v >> 1;
+    v |= v >> 2;
+    v |= v >> 4;
+    v |= v >> 8;
+    v |= v >> 16;
+    return v + 1;
+}
```
jt2594838, line 258: Is it faster to do so or store a pre-calculated array and use the number of leading zeros of n as the index?

cpp/src/common/global.cc
```diff
@@ -61,22 +51,24 @@ void init_config_value() {
     g_config_value_.boolean_encoding_type_ = PLAIN;
     g_config_value_.int32_encoding_type_ = TS_2DIFF;
     g_config_value_.int64_encoding_type_ = TS_2DIFF;
-    g_config_value_.float_encoding_type_ = GORILLA;
-    g_config_value_.double_encoding_type_ = GORILLA;
+    g_config_value_.float_encoding_type_ = PLAIN;
+    g_config_value_.double_encoding_type_ = PLAIN;
     g_config_value_.string_encoding_type_ = PLAIN;
```
jt2594838, line 66: Why is this?

cpp/src/common/global.cc
```diff
@@ -61,22 +51,24 @@ void init_config_value() {
...
-#ifdef ENABLE_LZ4
+    // Pick the strongest compressor that was actually compiled in. Gating on
+    // ENABLE_LZ4 while setting SNAPPY (the original code) would request a
+    // compressor that the factory can't produce when the build disables
+    // Snappy, returning nullptr at write time.
+#ifdef ENABLE_SNAPPY
+    g_config_value_.default_compression_type_ = SNAPPY;
+#elif defined(ENABLE_LZ4)
     g_config_value_.default_compression_type_ = LZ4;
 #else
```
jt2594838: Why is SNAPPY stronger than LZ4?

cpp/src/common/path.h
```diff
@@ -31,9 +36,57 @@ struct Path {
...
+                        std::make_shared<StringArrayDeviceID>(device_joined);
+                    measurement_ = nodes[nodes.size() - 1];
+                    full_path_ =
+                        device_id_->get_device_name() + "." + measurement_;
+                } else {
+                    full_path_ = path_sc;
+                    device_id_ = std::make_shared<StringArrayDeviceID>();
+                    measurement_ = path_sc;
+                }
+            }
```
jt2594838, line 86: Is this join necessary?

Why is device_id empty in the else branch?

cpp/src/common/statistic.h
```diff
@@ -634,13 +684,41 @@ class Int32Statistic : public Statistic {
...
+        for (uint32_t i = start; i < count; i++) {
+            if (timestamps[i] < start_time_) start_time_ = timestamps[i];
+            if (timestamps[i] > end_time_) end_time_ = timestamps[i];
+            if (values[i] < min_value_) min_value_ = values[i];
+            if (values[i] > max_value_) max_value_ = values[i];
+            sum_value_ += (int64_t)values[i];
+        }
+        last_value_ = values[count - 1];
+        count_ += (count - start);
+    }
```
jt2594838, line 720: If the timestamps are sorted, there is no point in updating start_time_ and end_time_ each time.

If not, first_value and last_value should be associated with the timestamp that is start_time or end_time.

cpp/src/common/tablet.cc
```diff
@@ -98,14 +100,8 @@ int Tablet::init() {
...
-                auto* sc = static_cast<StringColumn*>(common::mem_alloc(
-                    sizeof(StringColumn), common::MOD_TABLET));
-                if (sc == nullptr) return E_OOM;
-                new (sc) StringColumn();
-                // 8 bytes/row is a conservative initial estimate for short
-                // string columns (e.g. device IDs, tags). The buffer grows
-                // automatically on demand via mem_realloc.
-                sc->init(max_row_num_, max_row_num_ * 8);
+                auto* sc = new StringColumn();
+                sc->init(max_row_num_, max_row_num_ * 32);
```
jt2594838, line 109: Why is the mem control disabled?

cpp/src/common/tablet.cc
```diff
@@ -156,7 +153,7 @@ void Tablet::destroy() {
                 case TEXT:
                 case STRING:
                     value_matrix_[c].string_col->destroy();
-                    common::mem_free(value_matrix_[c].string_col);
+                    delete value_matrix_[c].string_col;
```
jt2594838: Explain this

cpp/src/common/tablet.cc
```diff
@@ -291,14 +306,74 @@ int Tablet::set_column_string_values(uint32_t schema_index,
...
+        sc->offsets[i] = i * str_len;
+        memcpy(sc->buffer + i * str_len, str, str_len);
+    }
+    sc->offsets[count] = total_bytes;
+    sc->buf_used = total_bytes;
+
+    bitmaps_[schema_index].clear_all();
     cur_row_size_ = std::max(count, cur_row_size_);
     return E_OK;
 }
```
jt2594838, line 364: May introduce RleStringColumn and some encapsulation

cpp/src/common/thread_pool.h
```diff
@@ -98,7 +113,23 @@ class ThreadPool {
...
+            // contract of the public submit(std::function<void()>) overload
+            // which has no way to surface the failure back to the caller.
+            // submit<F>() callers receive their error via the std::future
+            // wrapper installed by std::packaged_task — that path never
+            // reaches here, so this catch only fires for fire-and-forget
+            // tasks where the alternative is termination.
+            try {
+                task();
+            } catch (...) {
+                // Intentionally suppressed; see comment above.
```
jt2594838: Can we at least print some logs?

cpp/src/compress/uncompressed_compressor.h
```diff
@@ -43,11 +62,33 @@ class UncompressedCompressor : public Compressor {
...
+            common::mem_alloc(compressed_buf_len, common::MOD_COMPRESSOR_OBJ));
+        if (buf == nullptr) {
+            return common::E_OOM;
+        }
+        memcpy(buf, compressed_buf, compressed_buf_len);
+        uncompressed_buf = buf;
+        uncompressed_buf_ = buf;
         uncompressed_buf_len = compressed_buf_len;
         return common::E_OK;
     }
```
jt2594838, line 86: Is this really necessary?

cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -706,998 +744,1053 @@ DeviceSchema* tsfile_reader_get_all_timeseries_schemas(TsFileReader reader,
...
+        storage::TagFilterBuilder builder(schema.get());                      \
+        return builder.method(column_name, value);                            \
+    }
+
+DEFINE_TAG_FILTER_FACTORY(eq, eq)
+DEFINE_TAG_FILTER_FACTORY(neq, neq)
+DEFINE_TAG_FILTER_FACTORY(lt, lt)
+DEFINE_TAG_FILTER_FACTORY(lteq, lteq)
+DEFINE_TAG_FILTER_FACTORY(gt, gt)
+DEFINE_TAG_FILTER_FACTORY(gteq, gteq)
```
jt2594838: If all are the same, is it necessary to define two args？

cpp/src/common/global.cc
```diff
@@ -61,22 +51,24 @@ void init_config_value() {
...
     g_config_value_.int64_encoding_type_ = TS_2DIFF;
-    g_config_value_.float_encoding_type_ = GORILLA;
-    g_config_value_.double_encoding_type_ = GORILLA;
+    g_config_value_.float_encoding_type_ = PLAIN;
+    g_config_value_.double_encoding_type_ = PLAIN;
     g_config_value_.string_encoding_type_ = PLAIN;
-    // Default compression type is LZ4
-#ifdef ENABLE_LZ4
+    // Pick the strongest compressor that was actually compiled in. Gating on
+    // ENABLE_LZ4 while setting SNAPPY (the original code) would request a
```
ColinLeeo: should revert

cpp/src/encoding/decoder.h
```diff
@@ -37,6 +38,140 @@ class Decoder {
...
+    // Returns true if a block was peeked; false if not supported or no data.
+    // After peeking, caller must either:
+    //   - Call skip_peeked_block_int64() to skip the block
+    //   - Call read_batch_int64() which will use the peeked header
+    virtual bool peek_next_block_range_int64(common::ByteStream& in,
+                                             int64_t& block_min,
+                                             int64_t& block_max,
+                                             int& block_count) {
+        return false;
+    }
```
jt2594838: What is a block defined as, and how is it divided?
Why is there only int64?
What does `block_count` refer to?

cpp/src/encoding/int32_sprintz_decoder.h
```diff
@@ -125,9 +125,8 @@ class Int32SprintzDecoder : public SprintzDecoder {
             decode_size_ = bit_width_ & ~(1 << 7);
             Int32RleDecoder decoder;
             for (int i = 0; i < decode_size_; ++i) {
-                if (RET_FAIL(decoder.read_int(current_buffer_[i], input))) {
-                    return ret;
-                }
+                int ret = decoder.read_int(current_buffer_[i], input);
+                if (ret != common::E_OK) return ret;
```
jt2594838: Explain this change

cpp/src/encoding/ts2diff_decoder.h
```diff
@@ -274,6 +463,436 @@ inline int64_t TS2DIFFDecoder<int64_t>::decode(common::ByteStream& in) {
...
+    bits_left_ = 0;
+    buffer_ = 0;
+
+    block_min = first_value_;
+    block_count = write_index_ + 1;
+
+    // Look-ahead: since timestamps are monotonically increasing, the true
+    // block_max is the last timestamp, which equals next block's first_value_.
+    // The next block header starts at read_pos + packed_bytes. first_value_ is
+    // at offset 16 within the header
```
jt2594838, line 816: What if TS_2Diff is employed to a non-time column?

cpp/src/file/tsfile_io_reader.cc
```diff
@@ -80,6 +85,95 @@ int TsFileIOReader::alloc_ssi(std::shared_ptr<IDeviceID> device_id,
...
+    std::shared_ptr<IDeviceID> device_id,
+    const std::vector<std::string>& measurement_names,
+    TsFileSeriesScanIterator*& ssi, common::PageArena& pa,
+    Filter* time_filter) {
+    int ret = E_OK;
+    if (RET_FAIL(load_tsfile_meta_if_necessary())) return ret;
+
+    ssi = new TsFileSeriesScanIterator;
+    ssi->init(device_id, measurement_names.empty() ? "" : measurement_names[0],
+              read_file_, time_filter, pa);
```
jt2594838: No mem-control here?

cpp/src/file/tsfile_io_reader.cc
```diff
@@ -80,6 +85,95 @@ int TsFileIOReader::alloc_ssi(std::shared_ptr<IDeviceID> device_id,
...
+    for (const auto& meas_name : measurement_names) {
+        std::shared_ptr<IMetaIndexEntry> meas_entry;
+        int64_t meas_end_offset = 0;
+        if (RET_FAIL(load_measurement_index_entry(
+                meas_name, top_node, meas_entry, meas_end_offset))) {
+            // Measurement not found — abort multi path
+            delete ssi;
+            ssi = nullptr;
+            return ret;
+        }
```
jt2594838: Can we retrieve all measurements in one iteration?

cpp/src/file/tsfile_io_reader.cc
```diff
@@ -323,44 +384,111 @@ int TsFileIOReader::load_tsfile_meta() {
...
-    void* m_idx_node_buf = pa.alloc(sizeof(MetaIndexNode));
-    if (IS_NULL(data_buf) || IS_NULL(m_idx_node_buf)) {
-        return E_OOM;
+    const int64_t read_size_i64 = end_offset - start_offset;
+    // read_file_->read() takes int32_t; a meta index node larger than 2 GiB
+    // is implausible but explicitly reject it instead of silently truncating
+    // the read length and corrupting the parse.
+    if (read_size_i64 <= 0 || read_size_i64 > INT32_MAX) {
+        return false;
     }
```
jt2594838, line 457: This is not explicit enough, better to return a clearer code.

cpp/src/file/tsfile_io_reader.cc
```diff
@@ -411,12 +541,15 @@ int TsFileIOReader::load_device_index_entry(
...
         return E_DEVICE_NOT_EXIST;
     }
     auto index_node = it->second;
+    if (index_node == nullptr) {
+        return E_DEVICE_NOT_EXIST;
+    }
     if (index_node->node_type_ == LEAF_DEVICE) {
+        // FIXME
         ret = index_node->binary_search_children(
             device_name, true, device_index_entry, end_offset);
```
jt2594838: Fix what

cpp/src/file/tsfile_io_writer.h
```diff
@@ -214,6 +222,9 @@ class TsFileIOWriter {
     ChunkGroupMeta* cur_chunk_group_meta_;
     int32_t chunk_meta_count_;  // for debug
     common::SimpleList<ChunkGroupMeta*> chunk_group_meta_list_;
+    // O(1) lookup for existing ChunkGroupMeta by device name, avoiding the
+    // O(N) linear scan through chunk_group_meta_list_ per device.
+    std::unordered_map<std::string, ChunkGroupMeta*> chunk_group_meta_index_;
```
jt2594838: What if two devices share the same string name, like ("a.", "b") and ("a", ".b").

And what if one device have two chunk groups?

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -55,6 +65,25 @@ int SingleDeviceTsBlockReader::init(DeviceQueryTask* device_query_task,
...
+            nchunks++;
+            if (it.get()->statistic_) {
+                total += it.get()->statistic_->count_;
+            }
+        }
+        if (total == 0 && nchunks == 1 && fallback != nullptr) {
+            total = fallback->count_;
+        }
+        return total;
+    };
```
jt2594838, line 86: chunk_count -> point_count/count_chunk_point

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -149,32 +181,198 @@ int SingleDeviceTsBlockReader::init_internal(DeviceQueryTask* device_query_task,
...
+        bool all_outside = true;
+        for (const auto* ts_idx : time_series_indexs) {
+            if (ts_idx == nullptr) continue;
+            auto* chunk_list = ts_idx->is_aligned()
+                                   ? ts_idx->get_time_chunk_meta_list()
+                                   : ts_idx->get_chunk_meta_list();
+            if (chunk_list == nullptr) {
+                all_outside = false;
+                break;
+            }
```
jt2594838, line 207: Why break when any timeseries does not have chunks?

cpp/src/reader/block/single_device_tsblock_reader.cc
```diff
@@ -149,32 +181,198 @@ int SingleDeviceTsBlockReader::init_internal(DeviceQueryTask* device_query_task,
...
-    if (dense_row_count_ >= 0 && remaining_offset_ >= dense_row_count_) {
-        remaining_offset_ -= dense_row_count_;
-        delete current_block_;
-        current_block_ = nullptr;
-        return common::E_OK;
+    // Fast path: when every aligned column is provably dense (same total row
+    // count across time + value chunks), bulk-copy from SSI tsblock to caller
+    // tsblock instead of per-row merging.  compute_dense_row_count() returns
+    // -1 if the device is not provably dense, which gates safety.
+    const bool enable_dense_aligned_fast_path = true;
```
jt2594838: What is the point of defining a const here?

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -550,19 +666,19 @@ int AlignedChunkReader::decode_time_value_buf_into_tsblock(
...
-                    break;                                                     \
-                }                                                              \
                 ret = time_decoder_->read_int64(time, time_in);                \
                 if (ret != E_OK) {                                             \
                     break;                                                     \
                 }                                                              \
+                if (UNLIKELY(!row_appender.add_row())) {                       \
+                    ret = E_OVERFLOW;                                          \
+                    break;                                                     \
+                }                                                              \
```
jt2594838, line 684: Explain this

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -632,6 +748,566 @@ int AlignedChunkReader::i32_DECODE_TYPED_TV_INTO_TSBLOCK(
...
+            } else {
+                int32_t val = values[val_idx++];
+                if (filter != nullptr && !block_all_pass &&
+                    !filter->satisfy(times[i], (int64_t)val)) {
+                    continue;
+                }
+                if (UNLIKELY(!row_appender.add_row())) {
+                    ret = E_OVERFLOW;
+                    break;
+                }
```
jt2594838: Will the already-read values be missing in this case?

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -632,6 +748,566 @@ int AlignedChunkReader::i32_DECODE_TYPED_TV_INTO_TSBLOCK(
...
+        }
+        if (ret != E_OK) break;
+    }
+    return ret;
+}
+
+int AlignedChunkReader::double_DECODE_TV_BATCH(ByteStream& time_in,
+                                               ByteStream& value_in,
+                                               RowAppender& row_appender,
+                                               Filter* filter) {
```
jt2594838: Is it possible to use a template or extraction to reduce redundancy?

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -810,10 +1497,1572 @@ int AlignedChunkReader::get_next_page(TsBlock* ret_tsblock,
...
+    offset += sizeof(uint32_t);
+    uint32_t bitmap_bytes = (data_num + 7) / 8;
+    if (uncompressed_size - offset < bitmap_bytes) {
+        col->compressor->after_uncompress(uncompressed_buf);
+        return E_TSFILE_CORRUPTED;
+    }
+    pps.notnull_bitmap.resize(bitmap_bytes);
+    for (size_t i = 0; i < pps.notnull_bitmap.size(); i++) {
+        pps.notnull_bitmap[i] = *(uncompressed_buf + offset++);
+    }
```
jt2594838: memcpy?

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -810,10 +1497,1572 @@ int AlignedChunkReader::get_next_page(TsBlock* ret_tsblock,
...
+
+    // Dispatch:
+    //   - Single-thread (or thread pool disabled) → 4/6 thesis path:
+    //     per-page parallel decompress + serial batch decode+scatter via
+    //     multi_DECODE_TV_BATCH (stack-buffer based, no per-chunk allocation).
+    //   - Multi-thread with ≤6 value columns → chunk-level pre-decode + bulk
+    //     memcpy scatter.  Narrow chunks fit in cache and pay off the upfront
+    //     buffer allocation.
+    //   - Multi-thread with >6 value columns → 4/6 path; per_page_state would
+    //     thrash cache at high column count.
```
jt2594838: What is " 4/6 thesis path"

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -810,10 +1497,1572 @@ int AlignedChunkReader::get_next_page(TsBlock* ret_tsblock,
...
+                for (uint32_t c = 0; c < num_cols; c++) {
+                    auto* col = value_columns_[c];
+                    auto& pps = col->per_page_state[current_page_plan_index_];
+                    bool is_null = true;
+                    if (!pps.notnull_bitmap.empty()) {
+                        is_null =
+                            ((pps.notnull_bitmap[page_time_cursor_ / 8] &
+                              0xFF) &
+                             (null_mask_base >> (page_time_cursor_ % 8))) == 0;
+                    }
```
jt2594838, line 1992: An empty bitmap indicates all null?

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -810,10 +1497,1572 @@ int AlignedChunkReader::get_next_page(TsBlock* ret_tsblock,
...
+            // Value buffer for fixed-width types — up to 129 * 8 bytes
+            char val_buf[BATCH * 8];
+            int val_count;
+            // Variable-length values for STRING/TEXT/BLOB columns.  Only
+            // populated when the column's data_type_ is variable; their
+            // bufs are owned by the caller-provided PageArena.
+            std::vector<common::String> str_vals;
+        };
+        // Allocate on heap if many columns, stack for small counts
+        std::vector<ColBatch> col_batches(num_cols);
```
jt2594838: Where is the code associated with the comment?

cpp/src/reader/aligned_chunk_reader.cc
```diff
@@ -810,10 +1497,1572 @@ int AlignedChunkReader::get_next_page(TsBlock* ret_tsblock,
...
+            bool all_nonnull = true;
+            for (uint32_t c = 0; c < num_cols; c++) {
+                auto dt = value_columns_[c]->chunk_header.data_type_;
+                if (col_batches[c].nonnull_count != time_count ||
+                    dt == common::STRING || dt == common::TEXT ||
+                    dt == common::BLOB) {
+                    all_nonnull = false;
+                    break;
+                }
+            }
```
jt2594838: all_nonnull_and_fixed_size?

cpp/src/writer/value_page_writer.cc
```diff
@@ -54,11 +54,11 @@ int ValuePageData::init(ByteStream& col_notnull_bitmap_bs, ByteStream& value_bs,
     if (RET_FAIL(common::copy_bs_to_buf(col_notnull_bitmap_bs,
                                         uncompressed_buf_ + sizeof(size),
                                         col_notnull_bitmap_buf_size_))) {
-    } else if (value_buf_size_ > 0 && RET_FAIL(common::copy_bs_to_buf(
```
ColinLeeo, line 57: need to check.

cpp/src/writer/tsfile_table_writer.cc
```diff
@@ -66,21 +66,48 @@ int storage::TsFileTableWriter::write_table(storage::Tablet& tablet) const {
...
 
     auto schema_map = tablet.get_schema_map();
-    std::map<std::string, int> schema_map_;
+    std::map<std::string, int> new_schema_map;
     for (auto iter = schema_map.begin(); iter != schema_map.end(); iter++) {
-        schema_map_[to_lower(iter->first)] = iter->second;
+        new_schema_map[to_lower(iter->first)] = iter->second;
     }
-    tablet.set_schema_map(schema_map_);
+    tablet.set_schema_map(new_schema_map);
```
ColinLeeo, line 83: should handle.

cpp/src/writer/time_chunk_writer.h
```diff
@@ -73,29 +98,25 @@ class TimeChunkWriter {
     Statistic* get_chunk_statistic() { return chunk_statistic_; }
     FORCE_INLINE int32_t num_of_pages() const { return num_of_pages_; }
 
+    int64_t estimate_max_series_mem_size();
```
ColinLeeo: out of order.

cpp/src/common/statistic.h
```diff
@@ -22,12 +22,18 @@
...
 #include "common/allocator/alloc_base.h"
 #include "common/allocator/byte_stream.h"
 #include "common/db_common.h"
 
+#if defined(__ARM_NEON) || defined(__ARM_NEON__)
+#include <arm_neon.h>
+#define TSFILE_HAS_NEON 1
+#endif
+
 namespace storage {
```
ColinLeeo: need check

cpp/src/reader/tsfile_series_scan_iterator.cc
```diff
@@ -34,6 +46,12 @@ void TsFileSeriesScanIterator::destroy() {
         delete tsblock_;
         tsblock_ = nullptr;
     }
+#ifdef ENABLE_THREADS
+    if (decode_pool_ != nullptr) {
+        delete decode_pool_;
+        decode_pool_ = nullptr;
+    }
+#endif
```
jt2594838: Is the pool allocated for each TsFileSeriesScanIterator?

If someone reads multiple tsfiles concurrently, the thread overhead may be non-negligible.

It would be better to use a global pool

cpp/src/reader/tsfile_series_scan_iterator.cc
```diff
@@ -91,74 +113,104 @@ int TsFileSeriesScanIterator::get_next(TsBlock*& ret_tsblock, bool alloc,
...
+                    if (should_skip_aligned_chunk_by_offset(time_cm,
+                                                            value_cm)) {
+                        continue;
+                    }
+                    chunk_reader_->reset();
+                    if (RET_FAIL(chunk_reader_->load_by_aligned_meta(
+                            time_cm, value_cm))) {
+                    }
+                    break;
                 }
```
jt2594838, line 209: Is this path necessary now?
Can we treat this as a special case of multi-aligned?

cpp/src/reader/tsfile_series_scan_iterator.cc
```diff
@@ -225,4 +377,29 @@ TsBlock* TsFileSeriesScanIterator::alloc_tsblock() {
...
 }
 
-}  // end namespace storage
\ No newline at end of file
+TsBlock* TsFileSeriesScanIterator::alloc_tsblock_multi() {
+    auto* acr = static_cast<AlignedChunkReader*>(chunk_reader_);
+
+    // Time column
+    ColumnSchema time_cd("time", common::INT64, common::SNAPPY,
+                         common::TS_2DIFF);
```
jt2594838: Why are the encoding and compression hard-coded?

cpp/src/reader/tsfile_series_scan_iterator.cc
```diff
@@ -225,4 +377,29 @@ TsBlock* TsFileSeriesScanIterator::alloc_tsblock() {
...
+                              ch.compression_type_, ch.encoding_type_);
+        tuple_desc_.push_back(value_cd);
+    }
+
+    tsblock_ = new TsBlock(&tuple_desc_);
+    if (E_OK != tsblock_->init()) {
+        delete tsblock_;
+        tsblock_ = nullptr;
+    }
+    return tsblock_;
```
jt2594838: No memory control?

cpp/src/writer/chunk_writer.h
```diff
@@ -103,6 +103,68 @@ class ChunkWriter {
...
+                    batch_size))) {
+                return ret;
+            }
+            offset += batch_size;
+            if (RET_FAIL(seal_cur_page_if_full())) {
+                return ret;
+            }
+        }
+        return ret;
+    }
```
jt2594838, line 173: TODO: add size-based control to avoid extremely large pages.

cpp/src/writer/page_writer.h
```diff
@@ -150,10 +150,63 @@ class PageWriter {
...
+                                                 time_out_stream_))) {
+        } else if (RET_FAIL(value_encoder_->encode_string_batch(
+                       buffer, offsets, start_idx, count, value_out_stream_))) {
+            partial_failure_ = true;
+        } else {
+            for (uint32_t i = 0; i < count; i++) {
+                uint32_t idx = start_idx + i;
+                uint32_t len = offsets[idx + 1] - offsets[idx];
+                common::String val(buffer + offsets[idx], len);
+                statistic_->update(timestamps[i], val);
```
jt2594838, line 194: Add update_batch for string statistics?

cpp/src/writer/tsfile_writer.cc
```diff
@@ -683,14 +692,26 @@ int64_t TsFileWriter::calculate_mem_size_for_all_group() {
...
-        int64_t mem_size = calculate_mem_size_for_all_group();
+        // chunk-writer memory drops to ~0 after flush, but chunk metadata
+        // (ChunkMeta / ChunkGroupMeta / per-statistic PageArenas) keeps
+        // accumulating until end_file().  Wide-schema or many-flush
+        // workloads can pile up tens of MB of metadata that the old
+        // threshold check ignored entirely — flush would never fire even
+        // though total writer memory was well past chunk_group_size_threshold_.
+        int64_t chunk_size = calculate_mem_size_for_all_group();
+        int64_t meta_size = calculate_meta_mem_size();
+        int64_t mem_size = chunk_size + meta_size;
```
jt2594838, line 721: The Java edition has a mechanism that flushes ChunkMetadata into a temporary file and uses it to construct the index tree once the file is closed.
May also introduce that mechanism into Cpp.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -907,247 +957,109 @@ int TsFileWriter::write_tablet_aligned(const Tablet& tablet) {
...
+        restore_seal();
+        return ret;
     }
-
-    // Write values column-by-column and seal at recorded boundaries.
     ASSERT(value_chunk_writers.size() == tablet.get_column_count());
-    for (uint32_t col = 0; col < value_chunk_writers.size(); col++) {
-        ValueChunkWriter* value_chunk_writer = value_chunk_writers[col];
+    for (uint32_t c = 0; c < value_chunk_writers.size(); c++) {
+        ValueChunkWriter* value_chunk_writer = value_chunk_writers[c];
```
jt2594838, line 1021: Do not add these meaningless changes.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -1214,175 +1149,330 @@ int TsFileWriter::write_table(Tablet& tablet) {
...
+                            value_chunk_writers[k]->has_current_page_data()) {
+                            if (RET_FAIL(value_chunk_writers[k]
+                                             ->seal_current_page())) {
+                                unrecoverable_ = true;
+                                return ret;
+                            }
+                        }
                     }
+                    time_cur_points = 0;
                 }
```
jt2594838, line 1241: Why will a newly created device have points?

cpp/src/writer/tsfile_writer.cc
```diff
@@ -1214,175 +1149,330 @@ int TsFileWriter::write_table(Tablet& tablet) {
...
-                if ((r = tcw->seal_current_page()) != E_OK) return r;
-                seg_start = boundary;
-            }
-            if (seg_start < ei) {
-                r = time_write_column(tcw, tablet, seg_start, ei);
+            uint32_t page_remaining =
+                (initial_page_points > 0 &&
+                 initial_page_points < page_max_points)
+                    ? (page_max_points - initial_page_points)
+                    : page_max_points;
```
jt2594838, line 1284: If initial_page_points  >= page_max_points, remaining will be page_max_points?

cpp/src/writer/tsfile_writer.cc
```diff
@@ -1214,175 +1149,330 @@ int TsFileWriter::write_table(Tablet& tablet) {
...
             }
-            if (ret != E_OK) return ret;
+            if (ret != E_OK) {
+                // One task aborted mid-batch while others may have written
+                // all of their rows; the per-column row counts no longer
+                // line up.  Mark the writer unrecoverable so flush/close
+                // can't seal a corrupt aligned chunk group.
+                unrecoverable_ = true;
+                return ret;
+            }
```
jt2594838, line 1382: Is it possible to clean the device's data?

cpp/src/writer/tsfile_writer.cc
```diff
@@ -1214,175 +1149,330 @@ int TsFileWriter::write_table(Tablet& tablet) {
...
+                            vt.vcw, vt.col_idx, ctx.segments,
+                            ctx.initial_page_points))) {
+                        unrecoverable_ = true;
+                        return ret;
+                    }
+                }
+            }
+        }
+    } else {
+        int start_idx = 0;
```
jt2594838, line 1406: This method is too long, may extract some sub-methods.

cpp/src/writer/tsfile_writer.cc
```diff
@@ -1396,14 +1486,13 @@ std::vector<std::pair<std::shared_ptr<IDeviceID>, int>>
 TsFileWriter::split_tablet_by_device(const Tablet& tablet) {
     std::vector<std::pair<std::shared_ptr<IDeviceID>, int>> result;
 
-    if (tablet.id_column_indexes_.empty()) {
+    if (tablet.id_column_indexes_.empty() || tablet.single_device_) {
+        // No tag columns or caller guarantees single device — skip boundary
+        // detection entirely.
         auto sentinel = std::make_shared<StringArrayDeviceID>("last_device_id");
         result.emplace_back(std::move(sentinel), 0);
```
jt2594838, line 1516: Is it possible to make the sentiel a static member?

cpp/src/writer/tsfile_writer.cc
```diff
@@ -1679,8 +1769,161 @@ int TsFileWriter::write_typed_column(ValueChunkWriter* value_chunk_writer,
...
+    uint32_t count = end_idx - start_idx;
+    if (count == 0) return ret;
+
+    bool has_null = false;
+    if (col_notnull_bitmap.may_have_set_bits()) {
+        for (uint32_t r = start_idx; r < end_idx; r++) {
+            if (col_notnull_bitmap.test(r)) {
+                has_null = true;
+                break;
+            }
```
jt2594838, line 1676: Is it possible to add test_range_any and test_range_all for Bitmap?

cpp/src/writer/value_chunk_writer.cc
```diff
@@ -110,7 +110,7 @@ int ValueChunkWriter::seal_cur_page(bool end_chunk) {
                 /*stat*/ false, /*data*/ false);
             if (IS_SUCC(ret)) {
                 save_first_page_data(value_page_writer_);
-                value_page_writer_.clear_page_data();
+                // value_page_writer_.destroy_page_data();
```
jt2594838: Why commented？

cpp/test/common/thread_pool_test.cc
```diff
@@ -0,0 +1,67 @@
...
+
+// Regression: TsFileWriter::thread_pool_ reads write_thread_count_ from the
+// global config at construction time.  If a long-lived writer was created
+// before libtsfile_init() ran the value is zero, and the ThreadPool used to
+// silently accept submit() but block wait_all() forever (no worker, active_
+// never reaches 0).  The pool now normalizes zero to a single worker so
+// submitted work makes progress and tasks don't hang.
+TEST(ThreadPoolTest, ZeroThreadPoolStillExecutesAndDrains) {
+    common::ThreadPool pool(0);
+    EXPECT_GE(pool.num_threads(), static_cast<size_t>(1));
```
jt2594838, line 37: Or, maybe we can define that any submitted task will be executed in the submitter thread synchronously when size == 0?

cpp/src/reader/filter/or_filter.h
```diff
@@ -50,6 +50,18 @@ class OrFilter : public BinaryFilter {
...
+        bool mask_right[129];
+        left_->satisfy_batch_time(times, count, mask);
+        right_->satisfy_batch_time(times, count, mask_right);
+        int pass = 0;
+        for (int i = 0; i < count; ++i) {
+            mask[i] = mask[i] || mask_right[i];
+            if (mask[i]) ++pass;
+        }
+        return pass;
+    }
```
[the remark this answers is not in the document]
ColinLeeo: Addressed — the fixed 129-element stack buffer is now a 256-element inline buffer with a heap fallback for count > 256, so any batch size is safe.

cpp/src/reader/filter/and_filter.h
```diff
@@ -50,6 +50,18 @@ class AndFilter : public BinaryFilter {
...
+        bool mask_right[129];
+        left_->satisfy_batch_time(times, count, mask);
+        right_->satisfy_batch_time(times, count, mask_right);
+        int pass = 0;
+        for (int i = 0; i < count; ++i) {
+            mask[i] = mask[i] && mask_right[i];
+            if (mask[i]) ++pass;
+        }
+        return pass;
+    }
```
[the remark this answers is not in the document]
ColinLeeo: Addressed — the fixed 129-element stack buffer is now a 256-element inline buffer with a heap fallback for count > 256, so any batch size is safe.

cpp/src/encoding/dictionary_encoder.h
```diff
@@ -83,7 +83,7 @@ class DictionaryEncoder : public Encoder {
         if (entry_index_.count(value) == 0) {
             index_entry_.push_back(value);
             map_size_ = map_size_ + value.length();
-            entry_index_[value] = static_cast<int>(index_entry_.size()) - 1;
+            entry_index_[value] = entry_index_.size();
```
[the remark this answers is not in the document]
ColinLeeo: Fixed — assigns via an explicit local (const int new_idx = static_cast<int>(index_entry_.size()) - 1; entry_index_[value] = new_idx;), removing the unspecified-evaluation-order dependency.

cpp/src/common/allocator/byte_stream.h
```diff
@@ -290,7 +292,14 @@ class ByteStream {
...
-        page_size_ = buf_len;
+        // page_mask_ is used as a bitmask and only works correctly for
+        // power-of-2 page sizes. Round up to the next power-of-2 so that
+        // (read_pos_ & page_mask_) gives the correct within-page offset and
+        // the page-crossing check doesn't misfire on arbitrary buffer sizes.
+        uint32_t ps = 1;
+        while (ps < (uint32_t)buf_len) ps <<= 1;
+        page_size_ = ps;
+        page_mask_ = ps - 1;
         head_.store(&wrapped_page_);
```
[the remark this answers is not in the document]
ColinLeeo: Addressed — the BufferIterator last page now clamps its length to total_size_ - consumed_ (with a comment on the rounded wrapped-page case), so reads stay bounded by total_size_

cpp/src/file/tsfile_io_writer.cc
```diff
@@ -269,17 +246,19 @@ int TsFileIOWriter::end_file() {
...
     OFFSET_DEBUG("before end file");
+
     if (RET_FAIL(write_log_index_range())) {
         std::cout << "writer range index error, ret =" << ret << std::endl;
     } else if (RET_FAIL(write_file_index())) {
         std::cout << "writer file index error, ret = " << ret << std::endl;
     } else if (RET_FAIL(write_file_footer())) {
         std::cout << "writer file footer error, ret = " << ret << std::endl;
-    } else if (RET_FAIL(sync_file())) {
+    } else if (g_config_value_.sync_on_close_ && RET_FAIL(sync_file())) {
```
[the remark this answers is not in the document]
ColinLeeo: sync_on_close_ defaults to true (in-class default, not reset by init_config_value()); the durability trade-off is documented in config.h.

cpp/test/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 #[[
-Licensed to the Apache Software Foundation (ASF) under one
+Licensed to the Apache Software Foundation(ASF) under one
```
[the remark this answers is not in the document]
ColinLeeo: Reverted — the Apache Software Foundation (ASF) spacing and the License casing are restored.

cpp/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 #[[
-Licensed to the Apache Software Foundation (ASF) under one
+Licensed to the Apache Software Foundation(ASF) under one
```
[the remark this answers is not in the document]
ColinLeeo: Same sa above.

cpp/test/cwrapper/c_release_test.cc
```diff
@@ -52,8 +53,7 @@ TEST_F(CReleaseTest, TestCreateFile) {
 
     // Folder
     file = write_file_new("test/", &error_no);
-    ASSERT_TRUE(error_no == RET_FILRET_OPEN_ERR ||
-                error_no == RET_ALREADY_EXIST);
+    ASSERT_EQ(RET_FILRET_OPEN_ERR, error_no);
```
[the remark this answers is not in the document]
ColinLeeo: Restored the platform-tolerant disjunction RET_FILRET_OPEN_ERR || RET_ALREADY_EXIST.

cpp/test/reader/tree_view/tsfile_tree_query_by_row_test.cc
```diff
@@ -1320,7 +1112,7 @@ TEST_F(TreeQueryByRowTest, DISABLED_QueryByRowFasterThanManualNext) {
     write_test_file(devices, measurements, num_rows);
 
     const int num_iters = 5;
-    const double tolerance = 0.2;
+    const double tolerance = 0.05;
```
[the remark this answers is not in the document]
ColinLeeo: Tolerance set back to 0.5 (50%) with a comment, matching the table-view sibling test.

cpp/src/writer/value_chunk_writer.h
```diff
@@ -110,6 +109,68 @@ class ValueChunkWriter {
...
+            uint32_t cur_points = value_page_writer_.get_point_numer();
+            uint32_t page_remaining =
+                common::g_config_value_.page_writer_max_point_num_ - cur_points;
+            if (page_remaining == 0) {
+                if (RET_FAIL(seal_cur_page(false))) {
+                    return ret;
+                }
+                page_remaining =
+                    common::g_config_value_.page_writer_max_point_num_;
+            }
```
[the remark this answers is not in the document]
ColinLeeo: All four batch sites now seal when cur_points >= page_cap before the subtraction, so page_remaining can't wrap.

cpp/src/cwrapper/arrow_c.cc
```diff
@@ -881,21 +837,26 @@ int ArrowStructToTablet(const char* table_name, const ArrowArray* in_array,
...
-                    (dtype == common::INT64 || dtype == common::DOUBLE) ? 8 : 4;
-                const void* data =
-                    static_cast<const char*>(col_arr->buffers[1]) +
-                    off * elem_size;
-                uint8_t* null_bm = InvertArrowBitmap(
-                    validity, off, static_cast<uint32_t>(n_rows));
-                if (validity != nullptr && null_bm == nullptr) {
-                    delete tablet;
-                    return common::E_OOM;
+                // Invert Arrow bitmap (1=valid) to TsFile bitmap (1=null)
```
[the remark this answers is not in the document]
ColinLeeo: Now honors col_arr->offset; a new ArrowIsValid() applies the bit offset (and the time buffer adds ts_arr->offset), fixing sliced Arrow arrays.

cpp/src/common/tsfile_common.h
```diff
@@ -590,10 +594,8 @@ class AlignedTimeseriesIndex : public ITimeseriesIndex {
         return value_ts_idx_->get_measurement_name();
     }
     virtual common::TSDataType get_data_type() const {
-        return value_ts_idx_ == nullptr ? common::INVALID_DATATYPE
-                                        : value_ts_idx_->get_data_type();
+        return time_ts_idx_->get_data_type();
     }
-    virtual bool is_aligned() const { return true; }
     virtual Statistic* get_statistic() const {
         return value_ts_idx_->get_statistic();
```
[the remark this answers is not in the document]
ColinLeeo: Returns the value-column type now (with an explanatory comment), not the time column type.

cpp/src/common/tsfile_common.h
```diff
@@ -631,14 +674,13 @@ class TSMIterator {
     // timeseries measurenemnt chunk meta info
     // map <device_name, <measurement_name, vector<chunk_meta>>>
     std::map<std::shared_ptr<IDeviceID>,
-             std::map<common::String, std::vector<ChunkMeta*>>,
-             IDeviceIDComparator>
+             std::map<common::String, std::vector<ChunkMeta*>>>
         tsm_chunk_meta_info_;
```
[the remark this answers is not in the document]
ColinLeeo: Added IDeviceIDComparator so the map keys compare by device-id content.

cpp/src/common/tsfile_common.cc
```diff
@@ -103,13 +103,8 @@ int TSMIterator::init() {
...
-            auto& merged =
-                tsm_chunk_meta_info_[chunk_group_meta_iter_.get()->device_id_];
-            for (auto& m_entry : tmp) {
-                auto& vec = merged[m_entry.first];
-                vec.insert(vec.end(), m_entry.second.begin(),
-                           m_entry.second.end());
-            }
+            tsm_chunk_meta_info_[chunk_group_meta_iter_.get()->device_id_] =
+                tmp;
         }
```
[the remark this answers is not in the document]
ColinLeeo: Merges per-device entries via vec.insert(...) instead of overwriting.

cpp/src/writer/tsfile_table_writer.cc
```diff
@@ -66,21 +66,38 @@ int storage::TsFileTableWriter::write_table(storage::Tablet& tablet) const {
...
     }
-    tablet.set_table_name(to_lower(tablet.get_table_name()));
-    for (size_t i = 0; i < tablet.get_column_count(); i++) {
-        tablet.set_column_name(i, to_lower(tablet.get_column_name(i)));
-    }
+    if (!names_lowered_) {
+        tablet.set_table_name(to_lower(tablet.get_table_name()));
+        for (size_t i = 0; i < tablet.get_column_count(); i++) {
+            tablet.set_column_name(i, to_lower(tablet.get_column_name(i)));
+        }
```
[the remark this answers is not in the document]
ColinLeeo: The `names_lowered_` flag is gone — names are lowercased on every call (idempotent).

cpp/src/common/global.cc
```diff
@@ -61,22 +51,19 @@ void init_config_value() {
...
-    if (hw_cores == 0) hw_cores = 1;  // fallback if detection fails
-    g_config_value_.parallel_write_enabled_ = (hw_cores > 1);
-    g_config_value_.write_thread_count_ =
-        static_cast<int32_t>(std::min(hw_cores, 64u));
-    // Enforce aligned page size limits strictly by default.
-    g_config_value_.strict_page_size_ = true;
+    g_config_value_.parallel_read_enabled_ = true;
+    g_config_value_.parallel_write_enabled_ = true;
+    g_config_value_.read_thread_count_ = 4;
+    g_config_value_.write_thread_count_ = 6;
```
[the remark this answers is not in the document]
ColinLeeo: float/double reverted to GORILLA, compression now respects ENABLE_LZ4, and the thread count comes from config (not hardcoded)

cpp/src/file/tsfile_io_writer.cc
```diff
@@ -40,71 +42,46 @@ namespace storage {
...
+                     chunk_meta != iter.get()->chunk_meta_list_.end();
+                     chunk_meta++) {
+                    if (chunk_meta.get()) {
+                        chunk_meta.get()->statistic_->destroy();
+                    }
+                }
             }
-            chunk_meta.get()->statistic_->destroy();
         }
     }
```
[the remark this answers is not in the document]
ColinLeeo: No leak — `cur_chunk_meta_` is registered in `appended_chunk_metas_` at allocation (start_flush_chunk), and `destroy()` frees each appended statistic's PageArena.

cpp/src/common/global.cc
```diff
@@ -24,26 +24,16 @@
 #endif
```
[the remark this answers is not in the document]
ColinLeeo: Restored `<strings.h>`/`<string.h>` and the `util_define.h` Windows `strncasecmp` shim.

cpp/test/reader/table_view/tsfile_table_query_by_row_test.cc
```diff
@@ -652,15 +763,16 @@ TEST_F(TableQueryByRowTest, DenseSingleDeviceSsiLevelPushdown) {
 
 // Pushdown is faster than full query + manual next: queryByRow(offset, limit)
 // skips at device/SSI/Chunk level; old query then manual next decodes every
-// row. Timing tolerance 20% to allow measurement noise.
+// row. Timing tolerance 5% to allow measurement noise.
```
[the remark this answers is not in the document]
ColinLeeo: `tolerance = 0.2` now

cpp/src/common/allocator/byte_stream.h
ColinLeeo: Switched to `clz`+shift (`1u << (32 - __builtin_clz(n-1))`, `_BitScanReverse` on MSVC) — fewest ops, no data load.

cpp/src/common/global.cc
ColinLeeo: Fixed

cpp/src/common/global.cc
ColinLeeo: Fixed

cpp/src/common/statistic.h
ColinLeeo: Fixed

cpp/src/common/tablet.cc
ColinLeeo: Fixed

cpp/src/common/tablet.cc
ColinLeeo: Fixed

cpp/src/common/tablet.cc
ColinLeeo: Good Idea， and may tracking as a dedicated follow-up.

cpp/src/common/thread_pool.h
ColinLeeo: Fixed

cpp/src/cwrapper/tsfile_cwrapper.cc
ColinLeeo: Fixed

cpp/src/encoding/decoder.h
ColinLeeo: Fixed and add comment. 
    // TS_2DIFF stores values in self-contained "blocks": a header (value
    // count + per-delta bit width), then a min-delta and a first value,
    // then `count` bit-packed delta-of-deltas.  A page may hold several
    // sequential blocks; the boundary is set by the encoder's batch size.
    // This call peeks the next block's header (without consuming the packed
    // payload) and reports the block's value range [block_min, block_max]
    // and `block_count` — the number of values the block covers (first value
    // plus the packed deltas).  The caller then either:
    //   - Call skip_peeked_block_int64() to skip the whole block when a
    //     filter excludes [block_min, block_max], or
    //   - Call read_batch_int64(), which reuses the peeked header.
    //
    // Implemented only for INT64 because it targets the time column, which is
    // always INT64 and monotonically increasing.  Monotonicity is what makes
    // the range recoverable from the header alone (min = first value, max =
    // the block's last timestamp, obtained by looking ahead to the next
    // block's first value).  Non-monotonic columns (INT32 / value columns)
    // can't derive a range cheaply, so they fall back to this default and
    // decode normally.
    // Returns true if a block was peeked; false if not supported or no data.

cpp/src/encoding/int32_sprintz_decoder.h
ColinLeeo: Fixed

cpp/src/encoding/ts2diff_decoder.h
ColinLeeo: `peek_next_block_range_int64` is only ever called on `time_decoder_`

cpp/src/file/tsfile_io_reader.cc
ColinLeeo: Fixed

cpp/src/file/tsfile_io_reader.cc
ColinLeeo: Fixed

cpp/src/file/tsfile_io_reader.cc
ColinLeeo: Fixed

cpp/src/file/tsfile_io_writer.h
ColinLeeo: Good questions. The string-keyed index was actually dead code (written, never read) — the real lookup was an O(N) linear scan using `IDeviceID::operator==`, which is collision-safe. I've wired up an O(log N) lookup keyed by IDeviceID content (`std::map` + `IDeviceIDComparator`), so `("a.","b")` vs `("a",".b")` (which both render as `"a..b"` and would collide as strings) stay distinct.

cpp/src/reader/block/single_device_tsblock_reader.cc
ColinLeeo: Renamed the lambda `chunk_count` → `count_chunk_points` (it sums `statistic_->count_`, i.e. points, not chunks).

cpp/src/reader/block/single_device_tsblock_reader.cc
ColinLeeo: a column with no chunk metadata can't be proven outside the filter, so the device can't be skipped; the decision is final, hence the break.

cpp/src/reader/block/single_device_tsblock_reader.cc
ColinLeeo: Fixed and removed

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: reverted

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: No data loss — the batch loop returns `E_OVERFLOW` (and reads nothing) when `row_appender.remaining() < BATCH`, so a batch is never partially appended;

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: Fixed

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: Fixed

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: Fixed. Outdated comment.

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: Yes — an empty `notnull_bitmap` means the column had no data for that page (a missing/sparse aligned measurement), so every row is null. Added a comment.

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: Fixed

cpp/src/reader/tsfile_series_scan_iterator.cc
ColinLeeo: Fixed

cpp/src/reader/tsfile_series_scan_iterator.cc
ColinLeeo: Do it in a follow-up pr.

cpp/src/reader/tsfile_series_scan_iterator.cc
ColinLeeo: Fixed， use INVALID as a placeholder.

cpp/src/writer/chunk_writer.h
ColinLeeo: Good idea.

cpp/src/writer/tsfile_writer.cc
ColinLeeo: Good idea, and may be introduced in the feature.

cpp/src/writer/page_writer.h
ColinLeeo: A string `update_batch` would mostly complete the API — string stats don't get the SIMD speedup the numeric batch path does (it'd loop internally), saving only N→1 virtual calls. Happy to add it for consistency; left as a low-priority follow-up for now.

cpp/src/writer/tsfile_writer.cc
ColinLeeo: `device_ctx_index` tracks devices seen in this tablet, but `do_check_schema_table` returns the device's persistent chunk writer, which can already hold points from earlier tablets/records in the same un-flushed chunk group — so `time_cur_points > 0` is expected on first sight.

cpp/src/writer/tsfile_writer.cc
ColinLeeo: the caller seals and resets `time_cur_points` to 0 the moment it reaches `page_max_points`, so `initial_page_points` is always in `[0, page_max_points)`. The `< page_max_points` guard is defensive; the `else` branch effectively serves `initial == 0` (a fresh page = full room).

cpp/src/reader/aligned_chunk_reader.cc
ColinLeeo: removed

cpp/src/writer/tsfile_writer.cc
ColinLeeo: Good idea, and may do this in a follow-up pr.

cpp/src/common/tablet.cc
```diff
@@ -156,6 +158,7 @@ void Tablet::destroy() {
                 case TEXT:
                 case STRING:
                     value_matrix_[c].string_col->destroy();
+                    value_matrix_[c].string_col->~StringColumn();
```
jt2594838: Is it possible to merge the two?
Their semantics are similar, using both may lead to confusion.

cpp/src/writer/tsfile_writer.cc
ColinLeeo: I'll do it as a follow-up.

cpp/src/writer/tsfile_writer.cc
ColinLeeo: I'll do it as a follow-up.

cpp/src/writer/value_chunk_writer.cc
ColinLeeo: It's intentional — `save_first_page_data()` transfers ownership of the page buffers to `first_page_data_` (shallow-copy + `release_cur_page_data()`), which is freed later via `free_first_writer_data()` once the deferred first page is written.

cpp/test/common/thread_pool_test.cc
ColinLeeo: Good idea, and I will do it in  the feature.

cpp/src/common/tablet.cc
ColinLeeo: Good idea, and fixed.

PR #824, opened 2026-05-26
Add Tablet.serializedSize() and comprehensive size validation tests.

luoluoyuyu:
Pre-allocate serialization buffer using exact size estimation, support OBJECT type in tablet serialize/deserialize path, and consolidate serializedSize tests.

### DIFF
--- a/java/tsfile/src/main/java/org/apache/tsfile/write/record/Tablet.java
+++ b/java/tsfile/src/main/java/org/apache/tsfile/write/record/Tablet.java
@@ -748,11 +748,24 @@ public class Tablet implements Accountable {
 
   /** Serialize {@link Tablet} */
   public ByteBuffer serialize() throws IOException {
-    try (PublicBAOS byteArrayOutputStream = new PublicBAOS();
+    final int serializedSize = serializedSize();
+    try (PublicBAOS byteArrayOutputStream = new PublicBAOS(serializedSize);
         DataOutputStream outputStream = new DataOutputStream(byteArrayOutputStream)) {
       serialize(outputStream);
       return ByteBuffer.wrap(byteArrayOutputStream.getBuf(), 0, byteArrayOutputStream.size());
     }
+  }
+
+  /** Return the exact serialized byte size of this tablet. */
+  public int serializedSize() {
+    int size = 0;
+    size += ReadWriteIOUtils.sizeToWrite(insertTargetName);
+    size += Integer.BYTES;
+    size += serializedSizeOfMeasurementSchemas();
+    size += serializedSizeOfTimes();
+    size += serializedSizeOfBitMaps();
+    size += serializedSizeOfValues();
+    return size;
   }
 
   public void serialize(DataOutputStream stream) throws IOException {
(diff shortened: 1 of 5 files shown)


Caideyipi:
I found a functional issue.

  `Tablet.serializedSize()` claims to return the exact serialized byte size, but it uses
  `ReadWriteIOUtils.sizeToWrite(insertTargetName)` to calculate string sizes. That helper uses `s.getBytes()`, which
  depends on the platform default charset. The actual serialization path uses `ReadWriteIOUtils.write(String, ...)`,
  which encodes strings with `TSFileConfig.STRING_CHARSET` (UTF-8).

  So when the device/table name, measurement name, or schema properties contain non-ASCII characters, `serializedSize()`
  can differ from the real serialized size if the process default charset is not UTF-8.

  This is probably not an issue when TsFile is used through IoTDB, because IoTDB startup sets the default charset. But
  TsFile can also be used independently, and in standalone usage this can make the size estimate incorrect and break the
  “exact size” guarantee.

  Suggested fix: make `ReadWriteIOUtils.sizeToWrite(String)` use `TSFileConfig.STRING_CHARSET`, consistent with the
  write path, and add a non-ASCII name test.

  There is also a CodeQL alert for integer narrowing/overflow in `serializedSizeOfTimes()`. Since this method is
  intended to return an exact byte size, that should probably be handled as well.

java/tsfile/src/main/java/org/apache/tsfile/write/record/Tablet.java
```diff
@@ -764,6 +777,107 @@ public void serialize(DataOutputStream stream) throws IOException {
...
+        if (bitMaps[i] == null || bitMaps[i].isAllUnmarked(rowSize)) {
+          size = Math.addExact(size, Byte.BYTES);
+        } else {
+          size = Math.addExact(size, Byte.BYTES);
+          size = Math.addExact(size, Integer.BYTES);
+          size =
+              Math.addExact(
+                  size,
+                  ReadWriteIOUtils.sizeToWrite(
+                      new Binary(bitMaps[i].getTruncatedByteArray(rowSize))));
```
jt2594838: This is too costly. May simply calculate from rowSize?

luoluoyuyu: fixed

java/tsfile/src/main/java/org/apache/tsfile/write/record/Tablet.java
```diff
@@ -811,11 +811,8 @@ private int serializedSizeOfBitMaps() {
         } else {
           size = Math.addExact(size, Byte.BYTES);
           size = Math.addExact(size, Integer.BYTES);
-          size =
-              Math.addExact(
-                  size,
-                  ReadWriteIOUtils.sizeToWrite(
-                      new Binary(bitMaps[i].getTruncatedByteArray(rowSize))));
+          size = Math.addExact(size, Integer.BYTES);
+          size = Math.addExact(size, BitMap.getSizeOfBytes(rowSize));
```
jt2594838, line 815: Why two integers？

luoluoyuyu: The two `Integer.BYTES` entries represent two different fields in the serialized format.

  In `writeBitMaps()`, a non-empty bitmap is serialized as:

  1. `hasBitMap` flag: 1 byte
  2. `rowSize`: 4 bytes
  3. `Binary` length prefix: 4 bytes
  4. bitmap bytes: `BitMap.getSizeOfBytes(rowSize)`

  The previous code used:

  ```java
  ReadWriteIOUtils.sizeToWrite(new Binary(bitMaps[i].getTruncatedByteArray(rowSize)))
  ```

  That includes the Binary length prefix plus the actual bitmap bytes.

  So the new code:
  ```java
  size = Math.addExact(size, Integer.BYTES); // rowSize
  size = Math.addExact(size, Integer.BYTES); // Binary length prefix
  size = Math.addExact(size, BitMap.getSizeOfBytes(rowSize)); // bitmap bytes
 ```

  is equivalent to the old calculation. The two integers are not duplicates: one is the bitmap logical size (rowSize), and the other is the length prefix
  written by ReadWriteIOUtils.write(Binary, stream).

PR #826, opened 2026-05-31
Fix typos in comments, docs, build scripts, and i18n messages

SpriCoder:
## Summary

Repository-wide typo cleanup found with `codespell` and triaged by hand.
**42 fixes across 38 files**, grouped into four commits (`cpp` / `java` /
`docs` / `build`). All changes are limited to comments, Javadoc,
documentation, log/message strings, CMake/`build.sh` messages, and one
file-local C++ variable. No program logic, public identifiers, or test
expectations were modified.

### Representative fixes
- **C++**: `carefull`→careful, `timestmap`→timestamp, `namepsace`→namespace, `diectory`→directory, `genereate`→generate, `usigned`→unsigned, and the file-local parameter `increament`→increment (6 uses in `byte_stream.h`).
- **Java**: `caculate`→calculate (RLE/Zigzag encoders), `colum`→column, `differnt`→different, `perpare`→prepare, `vlaue`→value, `raed`→read, `fot`→for, `pricipal`→principal (Javadoc).
- **i18n** (`messages.properties`): `occured`→occurred and a duplicated `with with` in Sprintz log messages. Only **values** changed; keys are untouched, so the en/zh key-set alignment enforced by `MessagesTest` still holds.
- **Docs**: `aquisition`→acquisition, `severval`→several, `Paritial`→Partial, `wirh`→with, `memroy`→memory. In the Chinese committers doc, `Committe`→`Committer` (the Apache role) — this also repairs a broken anchor so it points to `About.md#committer`, matching the existing `### Committer` heading.
- **Build**: `methodes`→methods, `swtich`→switch in `pom.xml` JaCoCo comments.

### Intentionally NOT changed
- `QueryExpression::destory()` in `cpp/src/reader/expression.{cc,h}` — it is a **public API method name**; renaming would be a breaking change and belongs in a dedicated refactor.
- i18n `supproted` / `greather` — explicitly preserved for backward compatibility (documented inline in the resource bundle).
- `bais` (= `ByteArrayInputStream` convention), `notIn` / `testIn` (API and test method names) — identifiers, not typos.
- `external/commons/collections4/*` (`overriden`, `removeable`, `tothe`) — vendored Apache Commons code, left as-is to ease upstream syncing.

## Test plan
- [ ] `./mvnw spotless:check` (formatting unaffected — comment/string-only edits)
- [ ] `./mvnw clean verify -P with-java` (sanity; `MessagesTest` should still pass since only message values changed)
- [ ] `./mvnw clean verify -P with-cpp` (comment/variable-only changes; build unaffected)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -187,7 +187,7 @@
 * Added accountable function to measurementSchema by @Caideyipi in #509
 * Correct the retained size calculation for BinaryColumn and BinaryColumnBuilder by @JackieTien97 in #514
 * add switch to disable native lz4 (#480) by @jt2594838 in #515
-* Correct the memroy calculation of BinaryColumnBuilder by @JackieTien97 in #530
+* Correct the memory calculation of BinaryColumnBuilder by @JackieTien97 in #530
 * Fetch max tsblock line number each time from TSFileConfig by @JackieTien97 in #535
 * Support set default compression by data type & Bump org.apache.commons:commons-lang3 from 3.15.0 to 3.18.0 by @jt2594838 in #547
 * Avoid calculating shallow size of map by @shuwenwei in #566
--- a/cpp/build.sh
+++ b/cpp/build.sh
@@ -149,7 +149,7 @@ then
   cd build/minsizerel
 else
   echo ""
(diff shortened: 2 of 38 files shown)


ColinLeeo:
LGTM

Young-Leo (approved):
LGTM!

PR #829, opened 2026-06-03
Add tsfile-cli: inspect and import .tsfile from the command line

SpriCoder:
## Summary
Adds `tsfile-cli`, a single pipe-friendly C++ command-line tool (`cpp/tools/`) for working with `.tsfile` files from the shell. Built entirely on the existing `storage::TsFileReader` / `TsFileTableWriter` APIs; no storage-engine changes.

### Inspect / export (read-only)
- `ls` — list devices (tree model) or tables (table model)
- `schema` — per-series datatype / encoding / compression
- `meta` — file-level summary: model, counts, global time range, file size
- `stats` — per-series `count, start, end, min, max, first, last, sum` (from statistics, no page scan)
- `count` — per-series row counts + total
- `head` / `cat` — preview / stream rows, with projection (`-m`), time range (`--start/--end`), `--offset` / `-n`
- `sample` — deterministic reservoir sample (`--seed`)
- Output `csv|tsv|json|table` (TTY-adaptive); data → stdout, diagnostics → stderr; exit codes `0/1/2/3`

### Import (write)
- `write` — import CSV/TSV rows into a new table-model `.tsfile`
- Explicit `--columns name:TYPE:tag|field` schema (no type inference); first input column is the timestamp
- Reads a file or stdin (`-`); `-o` output; `-f csv|tsv`; optional `--no-header` / `--header-match`
- Silent on success (`-v` prints a one-line summary), Unix-style

### Other
- `ReadFile::open` errors now go to stderr (were stdout) so read output stays pipe-clean
- Unit + in-process E2E tests: arg parsing, formatters, statistics, CSV/TSV parsing, and a write→read round-trip

## Test plan
- [x] `cd cpp && bash build.sh -t=Debug` builds `bin/tsfile-cli` + `TsFile_Test` (add `--disable-antlr4` on CMake >= 4)
- [x] CLI suites pass: `InputFormatTest`, `ParseArgsTest`, `RunCliTest`, `CliE2E`, `RowWriterTest`, `StatTableTest`
- [x] `tsfile-cli meta|ls|schema|stats|count <file.tsfile>`
- [x] round-trip: `printf 'time,id,v\n0,d,1\n' | tsfile-cli write --table t --columns "id:STRING:tag,v:INT64:field" -o out.tsfile -` then `tsfile-cli count -f tsv out.tsfile`


cpp/tools/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+if (ENABLE_ANTLR4)
+    target_include_directories(tsfile_cli_obj PUBLIC
+            ${PROJECT_SOURCE_DIR}/third_party/antlr4-cpp-runtime-4/runtime/src)
+endif ()
+
+target_compile_definitions(tsfile_cli_obj PRIVATE
+        TSFILE_CLI_VERSION="${TsFile_CPP_VERSION}")
+
+add_executable(tsfile_cli tools_main.cc $<TARGET_OBJECTS:tsfile_cli_obj>)
+target_include_directories(tsfile_cli PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
```
ColinLeeo, line 42: include_directories(
        ${LIBRARY_INCLUDE_DIR}
        ${THIRD_PARTY_INCLUDE}
)

do not use ${CMAKE_CURRENT_SOURCE_DIR} pls

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+        }
+        if (rc == kExitOk && writer->write_table(tablet) != 0) {
+            err << "Error: write_table failed\n";
+            rc = kExitRuntime;
+        }
+    }
+
+    if (rc == kExitOk) {
+        if (writer->flush() != 0 || writer->close() != 0) {
+            err << "Error: flush/close failed\n";
```
ColinLeeo: If an error occurs during the writing process, we should clean up the TsFile that was created but failed to be written.

ColinLeeo:
When handling exceptions in the CLI, we should return readable error messages instead of generic descriptions or error codes.

cpp/tools/format/result_set_format.cc
```diff
@@ -0,0 +1,178 @@
...
+        case common::INT64:
+        case common::TIMESTAMP:
+            ss << rs->get_value<int64_t>(i);
+            return ss.str();
+        case common::FLOAT:
+            ss << rs->get_value<float>(i);
+            return ss.str();
+        case common::DOUBLE:
+            ss << rs->get_value<double>(i);
+            return ss.str();
```
ColinLeeo, line 55: ss.str() here relies on ostringstream's default precision of 6 significant digits, which truncates FLOAT/DOUBLE values and prevents them from round-tripping losslessly through a pipe.

cpp/src/file/read_file.cc
```diff
@@ -49,9 +51,9 @@ int ReadFile::open(const std::string& file_path) {
     file_path_ = file_path;
     fd_ = ::open(file_path_.c_str(), O_RDONLY);
     if (fd_ < 0) {
-        std::cout << "open file " << file_path << "  error :" << fd_
+        std::cerr << "open file " << file_path << "  error :" << fd_
                   << std::endl;
-        std::cout << "open error" << errno << "  " << strerror(errno)
+        std::cerr << "open error" << errno << "  " << strerror(errno)
                   << std::endl;
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62. `ReadFile::open()` no longer prints `fd_` (always -1 on failure); it now reports `strerror(errno)` plus the numeric errno.

cpp/tools/cli/run_cli.cc
```diff
@@ -0,0 +1,215 @@
...
+    }
+    if (args.empty()) {
+        print_usage(err);
+        return kExitUsage;
+    }
+    if (p.command == "help" || p.command == "--help" || p.command == "-h" ||
+        (p.help && p.file.empty())) {
+        print_usage(out);
+        return kExitOk;
+    }
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62. Help now triggers whenever `-h/--help` is present regardless of a positional file, so `tsfile-cli <cmd> --help <file>` prints usage.

cpp/tools/cli/run_cli.cc
```diff
@@ -0,0 +1,215 @@
...
+namespace tsfile_cli {
+namespace {
+
+void print_usage(std::ostream& os) {
+    os << "Usage: tsfile-cli <command> [options] <file.tsfile>\n"
+          "Commands:\n"
+          "  ls       list devices (tree) or tables (table)\n"
+          "  schema   per-measurement data type/encoding/compression\n"
+          "  meta     file metadata summary\n"
+          "  stats    per-series row count and time range\n"
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62. The `stats` usage line now reads `per-series count, time range, min/max/first/last/sum`.

cpp/tools/commands/cmd_meta.cc
```diff
@@ -0,0 +1,52 @@
...
+
+namespace tsfile_cli {
+
+int cmd_meta(const ParsedArgs& args, storage::TsFileReader& reader,
+             OutputFormat fmt, std::ostream& out, std::ostream& /*err*/) {
+    RowWriter w(out, fmt,
+                {"file", "model", "version", "device_count", "table_count",
+                 "series_count", "start_time", "end_time", "bloom_filter",
+                 "file_size_bytes"},
+                {common::STRING, common::STRING, common::STRING, common::INT64,
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62 by dropping the `version` and `bloom_filter` columns. The public `TsFileReader` API exposes neither the file version nor the bloom filter, so rather than emit perpetually-null columns I removed them (docs updated to match).

cpp/tools/README.md
```diff
@@ -0,0 +1,209 @@
...
+Exit codes: `0` success, `1` usage/argument error, `2` file open/corrupt,
+`3` query/runtime error.
+
+### Reading
+
+| Command | Description |
+|---|---|
+| `ls` | List devices (tree model) or tables (table model), one name per line |
+| `schema` | Per-series `target, measurement, datatype, encoding, compression` |
+| `meta` | File summary: model, version, device/table/series counts, time range, file size |
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62: the README `meta` row no longer lists `version` (the column was removed from the command).

cpp/tools/skills/tsfile-cli/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Read
+
+`tsfile-cli <cmd> [opts] <file.tsfile>` · `tsfile-cli --help | --version | help <cmd>`
+
+| cmd | output | scans pages |
+|---|---|---|
+| `ls` | device (tree) / table (table) per line | no |
+| `schema` | `target,measurement,datatype,encoding,compression` | no |
+| `meta` | model, version, device/table/series counts, time range, bloom, size | no |
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62: the SKILL `meta` row no longer lists `version`/`bloom` (both columns were removed from the command).

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+                }
+            }
+        }
+    }
+
+    std::vector<DataRow> rows;
+    while (std::getline(*in, line)) {
+        ++line_no;
+        strip_cr(line);
+        if (line.empty()) {
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62. `write` now streams rows into fixed 1024-row `Tablet` batches and flushes as each batch fills, so memory stays flat regardless of input size — the full file is never buffered. Verified with a 2500-row import (3 batches) round-tripping to `count`=2500.

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+                return false;
+            }
+            tablet.add_value(row, def.name, v);
+            return true;
+        }
+        case common::INT32: {
+            long v = std::strtol(cell.c_str(), &e, 10);
+            if (e == nullptr || *e != '\0') {
+                error = "bad INT32 '" + cell + "'";
+                return false;
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62. Numeric parsing now clears `errno` and rejects `ERANGE` for INT32/INT64/FLOAT/DOUBLE; INT32 additionally range-checks against `INT32_MIN/MAX` (parsed via `strtoll`). The timestamp parse rejects `ERANGE` too. e.g. importing `3000000000` into an INT32 column now errors (`INT32 out of range`) instead of silently truncating.

cpp/tools/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+]]
+
+message("Running in tools directory")
```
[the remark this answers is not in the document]
SpriCoder: Fixed in 7bb6b62: removed the unconditional `message("Running in tools directory")`.

cpp/tools/format/output_format.cc
```diff
@@ -0,0 +1,321 @@
...
+    out_ << "\n";
+}
+
+void RowWriter::write(const std::vector<std::string>& cells,
+                      const std::vector<bool>& is_null) {
+    if (fmt_ == OutputFormat::kTable) {
+        rows_.push_back(cells);
+        rows_null_.push_back(is_null);
+        return;
+    }
```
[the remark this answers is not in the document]
SpriCoder: Left as-is by design. Table rendering must read all rows to compute column widths (same as `column -t` / `csvlook` / psql aligned mode), so it can't both stream and align. The unbounded/large path is already covered: when stdout is not a TTY, `resolve_format` returns **tsv**, which streams row-by-row with zero buffering — so piped exports never buffer. `table` is only the default on an interactive TTY, where output is bounded by the terminal anyway; a user dumping millions of rows would pass `-f tsv|csv`. Rearchitecting table output to stream isn't possible without dropping column alignment, so I'd prefer to keep the current behavior.

cpp/test/CMakeLists.txt
```diff
@@ -63,6 +63,26 @@ if (${DOWNLOADED})
...
+    # AppleClang searches /usr/local/include before CMake's generated -isystem
+    # paths. Force the vendored GTest headers ahead of any system installation.
+    foreach (GTEST_TARGET gtest gtest_main gmock gmock_main)
+        if (TARGET ${GTEST_TARGET})
+            set_target_properties(${GTEST_TARGET} PROPERTIES SYSTEM OFF)
+            target_include_directories(${GTEST_TARGET} BEFORE PRIVATE
+                    ${googletest_SOURCE_DIR}/googletest/include
+                    ${googletest_SOURCE_DIR}/googletest
+                    ${googletest_SOURCE_DIR}/googlemock/include
+                    ${googletest_SOURCE_DIR}/googlemock)
```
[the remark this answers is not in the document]
SpriCoder: Addressed in 7bb6b62. Removed the no-op `set_target_properties(... SYSTEM OFF)` (there is no boolean `SYSTEM` target property) and added the vendored GTest include dirs `BEFORE PRIVATE` to `TsFile_Test` itself, where the test sources are compiled and header resolution actually matters. The Apple `-iquote/-I` + `-std=c++14` options on the gtest targets remain since they govern GTest's own compilation. Verified the AppleClang Debug build still resolves the FetchContent 1.12.1 headers and all tool tests pass.

SpriCoder:
@ColinLeeo Addressed in 7bb6b62. The CLI now translates storage-engine error codes into readable phrases via a `query_error_text()` helper — e.g. `Error: query failed: table does not exist` instead of `query failed (code 49)` — used by both the row-query and `sample` paths, and the result-read failure path now prints a readable message too. The other diagnostics (file open/corrupt, bad field count, type/overflow, header mismatch) were already human-readable. If a specific path is still surfacing a raw code, point me at it and I'll map it.

cpp/tools/format/input_format.cc
```diff
@@ -0,0 +1,162 @@
...
+    }
+    return true;
+}
+
+std::vector<std::string> split_line(const std::string& line, char delim,
+                                    bool csv_quotes) {
+    std::vector<std::string> out;
+    std::string field;
+    if (!csv_quotes) {
+        for (char c : line) {
```
ColinLeeo: When handling STRING / TEXT columns, you may need to deal with embedded newlines inside quoted CSV fields.

ColinLeeo (changes requested):

ColinLeeo:
Good Job！

cpp/tools/cli/cli_args.cc
```diff
@@ -0,0 +1,203 @@
...
+                p.error = "Invalid --model: " + val + " (use tree|table)";
+                return p;
+            }
+            p.model = val;
+        } else if (a == "--no-header") {
+            p.no_header = true;
+        } else if (a == "-h" || a == "--help") {
+            p.help = true;
+        } else if (a == "--version") {
+            p.version = true;
```
jt2594838, line 196: May return earlier for these cases?

cpp/tools/cli/cli_args.h
```diff
@@ -0,0 +1,61 @@
...
+    bool no_header = false;
+    std::string model;
+    std::string output;
+    std::string columns;
+    bool verbose = false;
+    bool header_match = false;
+    bool help = false;
+    bool version = false;
+    std::string error;
+};
```
jt2594838, line 56: May add some comments.

cpp/tools/cli/run_cli.cc
```diff
@@ -0,0 +1,215 @@
...
+          "  sample   deterministic sample rows (use -n and --seed)\n"
+          "  write    import CSV/TSV rows into a new table tsfile "
+          "(--table, --columns, -o)\n"
+          "Options: -f/--format csv|tsv|json|table, -d/--device, -t/--table,\n"
+          "         -m/--measurements a,b, -n/--limit, --offset, --seed,\n"
+          "         --start, --end,\n"
+          "         --no-header, --model tree|table, -h/--help, --version\n"
+          "Write options: --table, --columns name:TYPE:tag|field,..., "
+          "-o/--output,\n"
+          "         --header-match, -v/--verbose\n";
```
jt2594838: May add some explaination for these options.

cpp/tools/cli/run_cli.cc
```diff
@@ -0,0 +1,215 @@
...
+}
+
+bool is_known_command(const std::string& c) {
+    static const std::set<std::string> kCmds = {"ls",    "schema", "meta",
+                                                "stats", "head",   "cat",
+                                                "count", "sample", "write"};
+    return kCmds.count(c) != 0;
+}
```
jt2594838, line 92: count or find?

cpp/tools/cli/run_cli.cc
```diff
@@ -0,0 +1,215 @@
...
+
+bool validate_write_flags(const ParsedArgs& p, std::ostream& err) {
+    if (p.table.empty()) {
+        err << "Error: write requires --table\n";
+        return false;
+    }
+    if (p.columns.empty()) {
+        err << "Error: write requires --columns\n";
+        return false;
+    }
```
jt2594838, line 130: Add abbreviations like:
-t
-col

cpp/tools/cli/run_cli.cc
```diff
@@ -0,0 +1,215 @@
...
+    if (p.format == ParsedArgs::Format::kJson ||
+        p.format == ParsedArgs::Format::kTable) {
+        err << "Error: write input format must be csv or tsv\n";
+        return false;
+    }
+    if (!p.measurements.empty() || !p.device.empty() || p.has_start ||
+        p.has_end || p.has_seed || p.limit != -1 || p.offset != 0) {
+        err << "Error: read-only flags are not valid for write\n";
+        return false;
+    }
```
jt2594838, line 148: Print these flags for the convenience of the user's self-check?

cpp/tools/commands/cmd_meta.cc
```diff
@@ -0,0 +1,52 @@
...
+    w.write({s.file, s.model, "", std::to_string(s.device_count),
+             std::to_string(s.table_count), std::to_string(s.series_count),
+             s.has_time_range ? std::to_string(s.start_time) : "",
+             s.has_time_range ? std::to_string(s.end_time) : "", "",
+             std::to_string(s.file_size_bytes)},
+            {false, false, true, false, false, false, !s.has_time_range,
+             !s.has_time_range, true, false});
+    w.finish();
+    return kExitOk;
+}
```
jt2594838, line 48: Seeing this, I hope someday we may move TsFileResource from IoTDB down to TsFile.

cpp/tools/commands/cmd_sample.cc
```diff
@@ -0,0 +1,112 @@
...
+                if (reader.get_timeseries_schema(did, sch) == 0) {
+                    for (auto& m : sch) {
+                        ms.push_back(m.measurement_name_);
+                    }
+                }
+            }
+            for (const std::string& m : ms) {
+                paths.push_back(dev + "." + m);
+            }
+        }
```
jt2594838: When devices are not provided, it would be better to use the interface that directly gets all metadata. 
Instead of querying by deviceId one by one.

cpp/tools/commands/cmd_schema.cc
```diff
@@ -0,0 +1,124 @@
...
+        if (kv.first) {
+            std::vector<storage::MeasurementSchema> ms;
+            if (reader.get_timeseries_schema(kv.first, ms) == 0) {
+                for (auto& m : ms) {
+                    enc_comp[m.measurement_name_] =
+                        std::make_pair(tsencoding_name(m.encoding_),
+                                       compression_name(m.compression_type_));
+                }
+            }
+        }
```
jt2594838, line 101: Why not use kv.second (or extend it)?

cpp/tools/commands/cmd_schema.cc
```diff
@@ -0,0 +1,124 @@
...
+        write_table_schema_rows(args, reader, w);
+        w.finish();
+        return kExitOk;
+    }
+
+    storage::DeviceTimeseriesMetadataMap meta =
+        reader.get_timeseries_metadata();
+    for (auto& kv : meta) {
+        std::string target = kv.first ? kv.first->get_device_name() : "";
+        if (!args.device.empty() && target != args.device) {
```
jt2594838, line 84: May support devices?

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+}
+
+bool add_typed_value(storage::Tablet& tablet, uint32_t row,
+                     const ColumnDef& def, const std::string& cell,
+                     std::string& error) {
+    if (cell.empty()) {
+        return true;  // null
+    }
+    char* e = nullptr;
+    switch (def.type) {
```
jt2594838, line 81: Some types are missing.

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+        fin.open(args.file.c_str());
+        if (!fin.is_open()) {
+            err << "Error: cannot open input: " << args.file << "\n";
+            return kExitFile;
+        }
+        in = &fin;
+    }
+
+    const char delim = (args.format == ParsedArgs::Format::kTsv) ? '\t' : ',';
+    const bool csv_quotes = (delim == ',');
```
jt2594838, line 207: Support user-defined delim?

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+                std::vector<std::string> h =
+                    split_line(line, delim, csv_quotes);
+                bool ok = (h.size() == columns.size() + 1);
+                for (size_t i = 0; ok && i < columns.size(); ++i) {
+                    if (h[i + 1] != columns[i].name) {
+                        ok = false;
+                    }
+                }
+                if (!ok) {
+                    err << "Error: header does not match --columns (line 1)\n";
```
jt2594838: Give the header or its position that does not match.

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+            ++line_no;
+            strip_cr(line);
+            if (args.header_match) {
+                std::vector<std::string> h =
+                    split_line(line, delim, csv_quotes);
+                bool ok = (h.size() == columns.size() + 1);
+                for (size_t i = 0; ok && i < columns.size(); ++i) {
+                    if (h[i + 1] != columns[i].name) {
+                        ok = false;
+                    }
```
jt2594838, line 246: break?

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+            err << "Error: bad timestamp '" << fields[0] << "' (line "
+                << line_no << ")\n";
+            return kExitRuntime;
+        }
+        DataRow r;
+        r.line_no = line_no;
+        r.timestamp = static_cast<int64_t>(ts);
+        r.cells.assign(fields.begin() + 1, fields.end());
+        rows.push_back(r);
+    }
```
jt2594838: Need some chunking to avoid OOM.

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+    std::vector<std::string> names;
+    std::vector<common::TSDataType> types;
+    std::vector<common::ColumnCategory> cats;
+    std::vector<common::ColumnSchema> col_schemas;
+    for (const ColumnDef& d : columns) {
+        names.push_back(d.name);
+        types.push_back(d.type);
+        cats.push_back(d.category);
+        col_schemas.push_back(common::ColumnSchema(
+            d.name, d.type, common::UNCOMPRESSED, common::PLAIN, d.category));
```
jt2594838: Better to use the default compression and encoding of each type.

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+
+    storage::WriteFile file;
+    int flags = O_WRONLY | O_CREAT | O_TRUNC;
+#ifdef _WIN32
+    flags |= O_BINARY;
+#endif
+    if (file.create(args.output, flags, 0666) != 0) {
+        err << "Error: cannot create output: " << args.output << "\n";
+        return kExitFile;
+    }
```
jt2594838, line 287: Give the error code?

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+    auto* schema = new storage::TableSchema(args.table, col_schemas);
+    auto* writer = new storage::TsFileTableWriter(&file, schema);
+
+    int rc = kExitOk;
+    const size_t kBatch = 1024;
+    for (size_t start = 0; start < rows.size() && rc == kExitOk;
```
jt2594838: May change rows to an iterator.

cpp/tools/commands/cmd_write.cc
```diff
@@ -0,0 +1,261 @@
...
+    flags |= O_BINARY;
+#endif
+    if (file.create(args.output, flags, 0666) != 0) {
+        err << "Error: cannot create output: " << args.output << "\n";
+        return kExitFile;
+    }
+    auto* schema = new storage::TableSchema(args.table, col_schemas);
+    auto* writer = new storage::TsFileTableWriter(&file, schema);
+
+    int rc = kExitOk;
```
jt2594838: return_code or result_code

cpp/tools/commands/row_query.cc
```diff
@@ -0,0 +1,111 @@
...
+                if (reader.get_timeseries_schema(did, sch) == 0) {
+                    for (auto& m : sch) {
+                        ms.push_back(m.measurement_name_);
+                    }
+                }
+            }
+            for (const std::string& m : ms) {
+                paths.push_back(dev + "." + m);
+            }
+        }
```
jt2594838: Similar problem here when devices are not provided.

cpp/tools/commands/statistics.cc
```diff
@@ -0,0 +1,215 @@
...
+
+namespace tsfile_cli {
+namespace {
+
+template <typename T>
+std::string value_to_string(T value) {
+    std::ostringstream ss;
+    ss << value;
+    return ss.str();
+}
```
jt2594838, line 39: Will this `T` introduce unnecessary copying?

cpp/tools/CMakeLists.txt
SpriCoder: Addressed in a8c38e3c. `tsfile_cli_obj` now includes library headers from the staged `${LIBRARY_INCLUDE_DIR}` + `${THIRD_PARTY_INCLUDE}` instead of reaching into `${PROJECT_SOURCE_DIR}/src`; `${CMAKE_CURRENT_SOURCE_DIR}` stays only for the tool's own `cli/ commands/ format/` headers. This matches what `test/CMakeLists.txt` already does (`include_directories(${LIBRARY_INCLUDE_DIR} ${THIRD_PARTY_INCLUDE})` for the library + `-I .../tools` for tool headers). Verified the Debug build still resolves all headers from the staged include tree.

cpp/tools/commands/cmd_write.cc
SpriCoder: Already handled as of 57392f0a (the strict-review follow-up). On any failure (`rc != kExitOk`) `cmd_write` now closes the `WriteFile` and calls `std::remove(args.output.c_str())` before returning, so a partial/corrupt .tsfile is never left behind. There's a regression test for it: `CliE2E.WriteFailureOnBadValueLeavesNoOutput` writes a bad value and asserts the output file does not exist afterward. Let me know if you'd prefer the cleanup factored differently.

cpp/tools/format/result_set_format.cc
SpriCoder: Fixed in a8c38e3c. `cell_to_string` now sets `std::setprecision(std::numeric_limits<float|double>::max_digits10)` for FLOAT (9 digits) and DOUBLE (17 digits) before streaming the value, so they round-trip losslessly. e.g. a FLOAT `0.1` now prints `0.100000001` and a DOUBLE `3.141592653589793` prints `3.1415926535897931` instead of the old 6-sig-digit `0.1` / `3.14159`. Added `CliE2E.WriteThenReadFloatDoubleRoundTripLossless` to lock this in.

cpp/tools/format/input_format.cc
SpriCoder: Fixed in a8c38e3c. CSV reads now go through a new `read_record()` helper instead of a bare `std::getline`: when a line opens a double-quote that isn't closed, it keeps consuming physical lines (preserving the newline) until the quote closes, so a quoted STRING/TEXT field can span multiple lines and still import as one row. It reports `lines_consumed` so line numbers in error messages stay accurate, and for TSV (`csv_quotes=false`) a quote is just data so no joining happens. Covered by `InputFormatTest.ReadRecordJoinsQuotedNewline` (+ CR-strip / TSV / unterminated-quote cases) and the e2e `CliE2E.WriteImportsQuotedFieldWithEmbeddedNewline`, which imports `"line one\nline two"` and asserts `count`=2 rows (not 3) with the newline preserved in the value.

cpp/tools/commands/statistics.cc
```diff
@@ -0,0 +1,215 @@
...
+    s.device_count = static_cast<long long>(reader.get_all_device_ids().size());
+    s.table_count =
+        static_cast<long long>(reader.get_all_table_schemas().size());
+    s.file_size_bytes = file_size(args.file);
+
+    ParsedArgs all = args;
+    all.device.clear();
+    all.table.clear();
+    all.measurements.clear();
+    std::vector<SeriesStatRow> rows = collect_series_stats(all, reader);
```
jt2594838, line 199: Why is the clear here?

cpp/tools/format/input_format.cc
```diff
@@ -0,0 +1,156 @@
...
+        out = common::DOUBLE;
+    } else if (u == "STRING") {
+        out = common::STRING;
+    } else if (u == "TEXT") {
+        out = common::TEXT;
+    } else {
+        return false;
+    }
+    return true;
+}
```
jt2594838: Should be a util of TsFile, the same below.
Missing some types.

cpp/tools/format/input_format.cc
```diff
@@ -0,0 +1,156 @@
...
+bool parse_category(const std::string& s, common::ColumnCategory& out) {
+    if (s == "tag") {
+        out = common::ColumnCategory::TAG;
+    } else if (s == "field") {
+        out = common::ColumnCategory::FIELD;
+    } else {
+        return false;
+    }
+    return true;
+}
```
jt2594838, line 85: Case sensitivity?

cpp/tools/format/input_format.cc
```diff
@@ -0,0 +1,156 @@
...
+
+bool parse_bool_cell(const std::string& s, bool& out) {
+    std::string l;
+    l.reserve(s.size());
+    for (char c : s) {
+        l += static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
+    }
```
jt2594838, line 136: Is there no standard method for this?

cpp/tools/format/output_format.cc
```diff
@@ -0,0 +1,321 @@
...
+            return OutputFormat::kJson;
+        case ParsedArgs::Format::kTable:
+            return OutputFormat::kTable;
+        case ParsedArgs::Format::kAuto:
+        default:
+            return stdout_is_tty ? OutputFormat::kTable : OutputFormat::kTsv;
+    }
+}
+
+const char* tsdatatype_name(common::TSDataType t) {
```
jt2594838, line 92: Add to TsFile common utils. The same below.

cpp/tools/format/output_format.cc
```diff
@@ -0,0 +1,321 @@
...
+        case common::BOOLEAN:
+        case common::INT32:
+        case common::INT64:
+        case common::FLOAT:
+        case common::DOUBLE:
+        case common::TIMESTAMP:
+            return true;
+        default:
+            return false;
+    }
```
jt2594838: Why define boolean as numeric?

cpp/tools/format/output_format.cc
jt2594838, line 279: Risk of OOM?

cpp/tools/format/result_set_format.cc
```diff
@@ -0,0 +1,178 @@
...
+        BufferedRow row = read_current_row(rs, types);
+        if (static_cast<long long>(reservoir.size()) < limit) {
+            reservoir.push_back(row);
+        } else {
+            std::uniform_int_distribution<long long> dist(0, seen);
+            long long idx = dist(rng);
+            if (idx < limit) {
+                reservoir[static_cast<size_t>(idx)] = row;
+            }
+        }
```
jt2594838, line 171: Will this introduce row copying? Better to avoid it.

cpp/tools/skills/tsfile-cli/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+| `stats` | per-series `count,start,end,min,max,first,last,sum` | no |
+| `count` | per-series counts + `total` row | no |
+| `head` | first N rows (default 10, `-n`) | yes |
```
jt2594838, line 49: It seems that for the table model, only the first N rows of the first table will be returned.

cpp/tools/cli/cli_args.cc
SpriCoder: Done in 032a5745. `--help`/`--version` now return immediately from the parse loop once seen, so the rest of the args are not parsed (help/version take precedence anyway in run_cli).

cpp/tools/cli/cli_args.h
SpriCoder: Done in 032a5745. Every field of `ParsedArgs` now has a trailing comment describing the flag it maps to.

cpp/tools/cli/run_cli.cc
SpriCoder: Done in 032a5745. `print_usage` now lists each option on its own line with a short explanation (format default, time bounds, projection, etc.).

cpp/tools/cli/run_cli.cc
SpriCoder: It is `count` (number of rows). The command counts rows per series and prints a `total` row; there is no `find`. Clarified the usage line to "number of rows (per series, plus a total)".

cpp/tools/cli/run_cli.cc
SpriCoder: `-t` already works as the short form of `--table` (parsed in cli_args.cc). `--columns` is write-only and has no short alias; the usage text now spells out the abbreviations alongside the long names.

cpp/tools/cli/run_cli.cc
SpriCoder: Done in 032a5745. `write -v` now echoes the resolved config to stderr before importing (table/output/input/format/header plus each parsed `name:TYPE:category`), so the user can self-check the flags.

cpp/tools/commands/cmd_meta.cc
SpriCoder: Agreed — that would let the CLI report richer file-level metadata. Filed as a follow-up; out of scope for this PR since TsFileResource lives in IoTDB today.

cpp/tools/commands/cmd_sample.cc
SpriCoder: Done in 032a5745. When no device is given, the tree path now collects every device/series from a single `get_timeseries_metadata()` call (shared `collect_tree_query_paths` helper) instead of a per-device `get_timeseries_schema` loop.

cpp/tools/commands/cmd_schema.cc
SpriCoder: `kv.second` (the TimeseriesMetadata list) exposes only measurement name + data type — not encoding/compression. Those two columns require a separate `get_timeseries_schema` call, which is why `enc_comp` is built. the data type itself does come from `kv.second`.

cpp/tools/commands/cmd_schema.cc
SpriCoder: Single-device filtering via `-d`/`-t` is supported; multi-device (a list) is not yet. Filed as a follow-up — for now schema without a filter already lists all devices.

cpp/tools/commands/cmd_write.cc
SpriCoder: Done in 032a5745. `write` now also accepts TIMESTAMP, DATE (YYYY-MM-DD), and BLOB (in addition to BOOLEAN/INT32/INT64/FLOAT/DOUBLE/STRING/TEXT). Covered by ParseColumnsSpecExtendedTypes and the WriteRoundTripsTimestampDateBlob e2e test.

cpp/tools/commands/cmd_write.cc
SpriCoder: The `-f csv|tsv` switch already covers the two common delimiters (comma/tab). An arbitrary user-defined delimiter is a separate feature; filed as a follow-up to avoid expanding the flag surface here.

cpp/tools/commands/cmd_write.cc
SpriCoder: Done in 032a5745. `--header-match` failures now name the offending position, e.g. "header column 2 is 'wrong', expected 's1'" (or the column-count mismatch). Covered by WriteHeaderMatchReportsMismatchPosition.

cpp/tools/commands/cmd_write.cc
SpriCoder: The header check now returns immediately on the first mismatching column (early `return kExitRuntime` instead of an `ok` flag), so it both breaks and reports the exact position.

cpp/tools/commands/cmd_write.cc
SpriCoder: Already handled as of 57392f0a: `write` streams rows into fixed 1024-row Tablet batches and flushes as each fills, so the full input is never buffered. Memory stays flat regardless of input size — verified by WriteStreamsLargeInputRoundTrips (3000 rows) and WriteRejectsOutOfOrderAcrossBatches.

cpp/tools/commands/cmd_write.cc
SpriCoder: Done in 032a5745. Each column's ColumnSchema now uses `common::get_value_encoder(type)` and `common::get_default_compressor()` (build-aware: LZ4 when enabled, else UNCOMPRESSED) instead of hard-coded PLAIN/UNCOMPRESSED. SchemaTableShowsEncodingAndCompression asserts e.g. INT64 -> TS_2DIFF.

cpp/tools/commands/cmd_write.cc
SpriCoder: Done in 032a5745. The writer failure messages now append the numeric code in addition to the readable phrase, e.g. "failed to write rows: ... (code N)" for write/flush/close.

cpp/tools/commands/cmd_write.cc
SpriCoder: Effectively done by the streaming rework (57392f0a): there is no longer a full `std::vector<DataRow> rows` buffered for the whole file — rows flow through a bounded 1024-row batch that is consumed and cleared as input is read, which is the iterator-like behavior you're after.

cpp/tools/commands/cmd_write.cc
SpriCoder: Done in 032a5745. Renamed `rc` to `result_code` throughout cmd_write.

cpp/tools/commands/row_query.cc
SpriCoder: Done in 032a5745, same fix as cmd_sample: row_query now calls the shared `collect_tree_query_paths`, which uses one whole-file `get_timeseries_metadata()` call when no device is given instead of a per-device loop.

cpp/tools/commands/statistics.cc
SpriCoder: Done in 032a5745. `value_to_string` now takes `const T&`. In practice it's only instantiated for scalar statistic values, but const-ref is the correct signature.

cpp/tools/commands/statistics.cc
SpriCoder: That clear is intentional: the `meta` summary describes the whole file, so the args copy's `-d`/`-t`/`-m` filters are cleared before collecting per-series stats — otherwise series_count and the overall time range would only reflect the filtered subset. Added a comment saying so (the file moved to statistics.cc).

cpp/tools/format/input_format.cc
SpriCoder: Done in 032a5745. The datatype-name parser now lives in common as `common::parse_data_type_name` (alongside `get_data_type_name`) and covers all named types incl. TIMESTAMP/DATE/BLOB; the CLI delegates to it.

cpp/tools/format/input_format.cc
SpriCoder: Type names were already case-insensitive; the column category (`tag`/`field`) now is too, via a tolower pass. Covered by ParseColumnsSpecCaseInsensitiveCategory.

cpp/tools/format/input_format.cc
SpriCoder: C++11/14 has no standard CSV/RFC-4180 parser, so this is a small hand-rolled splitter. It's now sunk into common as `common::split_csv_line` (shared by the CLI), with unit coverage in CsvUtilsTest.

cpp/tools/format/output_format.cc
SpriCoder: Done in 032a5745. `csv_escape` (and the CSV splitter) now live in `common/csv_utils.{h,cc}` as `common::csv_escape` / `common::split_csv_line`; the CLI delegates to them.

cpp/tools/format/output_format.cc
SpriCoder: Intentional: in JSON output, booleans are emitted as bare `true`/`false` (valid JSON literals), the same way numbers are emitted bare — only strings/blobs/dates get quoted. `emits_json_bare` groups BOOLEAN with the numeric types for that reason. CSV/TSV still print `true`/`false` text.

cpp/tools/format/output_format.cc
SpriCoder: The unbounded path is safe: when stdout is not a TTY, `resolve_format` returns tsv, which streams row-by-row with zero buffering — so piped/large exports never buffer. Table buffers all rows only to compute column widths (like `column -t`/psql), and it's the default only on an interactive TTY where output is bounded by the terminal; a user dumping millions of rows passes `-f tsv|csv`.

cpp/tools/format/result_set_format.cc
SpriCoder: Done in 032a5745. The reservoir now moves the BufferedRow (`std::move`) into the vector / over the replaced slot instead of copying.

cpp/tools/skills/tsfile-cli/SKILL.md
SpriCoder: Correct — for the table model, the row verbs without `-t` query only the first table. Documented that caveat in SKILL.md and README, and `-t <table>` targets a specific one. Multi-table fan-out is a possible follow-up.

cpp/tools/commands/cmd_schema.cc
SpriCoder: I think we may support multiple devices in later PR, bescause this pr is too big

cpp/tools/cli/run_cli.cc
SpriCoder: Follow-up in 6d30a8c7: the validation **error messages** now also show the short alias where one exists, e.g. `write requires -t/--table`, `-d/--device and -t/--table cannot be used together`, `-n/--limit must be >= -1`. Flags without a short form (`--columns`, `--offset`, etc.) stay long-only.

cpp/tools/format/output_format.cc
SpriCoder: The misleading `is_numeric` name was already renamed to `emits_json_bare` (that's why this thread shows Outdated) — it answers "does this type serialize to a bare/unquoted JSON token?", which BOOLEAN does (`true`/`false` are JSON literals). Added a comment in 02377a97 spelling that out so the BOOLEAN case no longer looks like it's calling boolean numeric.

cpp/tools/cli/run_cli.cc
jt2594838: I mean the set function used to judge whether the command exists.

cpp/tools/commands/row_query.cc
```diff
@@ -31,6 +32,57 @@
...
+        // A single device was requested: resolve just its series.
+        std::vector<std::string> ms = args.measurements;
+        if (ms.empty()) {
+            auto did =
+                std::make_shared<storage::StringArrayDeviceID>(args.device);
+            std::vector<storage::MeasurementSchema> sch;
+            if (reader.get_timeseries_schema(did, sch) == 0) {
+                for (auto& m : sch) {
+                    ms.push_back(m.measurement_name_);
+                }
```
jt2594838: May directly put into `paths` here

cpp/tools/commands/row_query.cc
```diff
@@ -31,6 +32,57 @@
...
+            }
+        }
+        for (const std::string& m : ms) {
+            paths.push_back(args.device + "." + m);
+        }
```
jt2594838: If measurements are provided, do not check whether they exist in this device?

jt2594838 (approved):

cpp/tools/commands/row_query.cc
SpriCoder: Done in 7bedf88. The intermediate `ms` vector is gone — the single-device branch now resolves the device's schema and pushes `device + "." + measurement` straight into `paths`.

cpp/tools/commands/row_query.cc
SpriCoder: Good catch — it didn't check before. Fixed in 7bedf88: the branch now always resolves the device's real schema and filters the provided measurements through it (the same `include_measurement` predicate the no-device path uses), so a measurement that isn't on the device is dropped instead of being queried blindly. `get_timeseries_schema` returns an empty list for a missing device, so an unknown device or an all-missing projection yields no paths and the caller reports `Error: no time series found`.

cpp/tools/cli/run_cli.cc
SpriCoder: Switched to `find` in e11c1c6: `return kCmds.find(c) != kCmds.end();`. It reads as an existence test rather than a count — equivalent for a `std::set` (unique keys, same O(log n)), and C++11-compatible since `contains()` is C++20.

ColinLeeo (approved):

cpp/tools/CMakeLists.txt
ColinLeeo: remove CMAKE_CURRENT_SOURCE_DIR

SpriCoder: Done in #835 — both `${CMAKE_CURRENT_SOURCE_DIR}` usages are replaced with `${CMAKE_SOURCE_DIR}/tools` (for `tsfile_cli_obj` and the `tsfile_cli` executable), matching the convention already used in `cpp/test/CMakeLists.txt`. Note the executable line is not redundant: `$<TARGET_OBJECTS:tsfile_cli_obj>` only pulls in object files and does not propagate the object library's PUBLIC include dirs, so `tools_main.cc` still needs `tools/` on its include path. Verified the Debug build links `tsfile_cli` and resolves all headers from the staged include tree.

PR #837, opened 2026-06-10
Fix several bugs in tsfile-cli and add tag filter commands

Alchuang22-dev:
## Summary

The PR is a plugin for #836.

**CLI install behavior.** `cpp/build.sh` no longer runs `make install` by default, which previously installed into CMake's default `/usr/local` prefix when users only intended to build. The script now builds only by default and runs install only when explicitly invoked with `install` / `--install`. `--help` and README docs were updated accordingly.

**Case-insensitive table filters.** Metadata commands now normalize table filter names before comparing against stored table schema names. This fixes `schema -t TABLE1`, `stats -t TABLE1`, and `count -t TABLE1` returning empty results for table-model files whose internal table names are lower-cased.

**Row-window pushdown.** `head` / `cat` now use `TsFileReader::queryByRow(...)` when `--offset` / `-n` can be pushed down, avoiding the previous behavior of scanning rows first and applying offset/limit in the formatter. Time-range queries (`--start` / `--end`) keep the existing path so offset/limit semantics remain "after time filtering" until `queryByRow` grows a time-range API.

**Tag filter predicates.** Added a simple table TAG predicate interface for row commands:
- `--tag-filter <column> <op> <value>`
- `--tag-between <column> <lower> <upper>`
- `--tag-not-between <column> <lower> <upper>`

Supported operators match the existing C++ tag filter builder: `eq`, `neq`, `lt`, `lteq`, `gt`, `gteq`, `regexp`, and `not-regexp`. The CLI constructs the underlying `TagFilterBuilder` filter and passes it through the reader's `tag_filter` path for `head`, `cat`, and `sample`.

**Docs and agent skill.** Updated `cpp/tools/README.md` and the in-repo `tsfile-cli` skill so users and agents can discover the explicit install mode, row-window behavior, and tag predicate options.

## Test plan

- [x] `cmake --build cpp/build/Debug --target tsfile_cli`
- [x] `cmake -S cpp -B cpp/build/Debug -DCMAKE_BUILD_TYPE=Debug -DBUILD_TEST=ON -DENABLE_ANTLR4=OFF`
- [x] `cmake --build cpp/build/Debug --target TsFile_Test`
- [x] `cpp/build/Debug/test/lib/TsFile_Test --gtest_filter='ParseArgsTest.*'`
- [x] `cpp/build/Debug/test/lib/TsFile_Test --gtest_filter='CliE2E.*'`
- [x] Regression coverage for case-insensitive table metadata filters: `schema`, `stats`, `count`
- [x] Regression coverage for `head` / `cat --offset/-n` row-window behavior, including time-range fallback semantics
- [x] Regression coverage for table TAG filters on `cat`, `head`, and `sample`, including invalid field-column rejection
- [x] Manual large-file-path check with `large_data.tsfile` for `cat/head --offset -n`
- [x] Manual tag-filter check with `large_data.tsfile`, including `eq`, `between`, and `regexp` predicates
- [x] `bash -n build.sh`
- [x] `bash build.sh --help`
- [x] `bash build.sh -t=Debug --disable-antlr4` confirms default build skips install

### DIFF
--- a/cpp/tools/commands/statistics.cc
+++ b/cpp/tools/commands/statistics.cc
@@ -27,6 +27,7 @@
 #include "commands/commands.h"
 #include "common/statistic.h"
 #include "reader/tsfile_reader.h"
+#include "utils/storage_utils.h"
 
 namespace tsfile_cli {
 namespace {
@@ -138,6 +139,7 @@ StatisticCells statistic_value_cells(storage::Statistic* st) {
 std::vector<SeriesStatRow> collect_series_stats(const ParsedArgs& args,
                                                 storage::TsFileReader& reader) {
     std::vector<SeriesStatRow> rows;
+    const std::string table_filter = storage::to_lower(args.table);
     storage::DeviceTimeseriesMetadataMap meta =
         reader.get_timeseries_metadata();
     for (auto& kv : meta) {
@@ -145,8 +147,8 @@ std::vector<SeriesStatRow> collect_series_stats(const ParsedArgs& args,
         if (!args.device.empty() && target != args.device) {
             continue;
         }
-        if (!args.table.empty() && kv.first &&
-            kv.first->get_table_name() != args.table) {
+        if (!table_filter.empty() && kv.first &&
+            kv.first->get_table_name() != table_filter) {
             continue;
         }
         for (auto& ts : kv.second) {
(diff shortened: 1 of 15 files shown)


cpp/tools/commands/statistics.cc
```diff
@@ -138,15 +139,16 @@ StatisticCells statistic_value_cells(storage::Statistic* st) {
 std::vector<SeriesStatRow> collect_series_stats(const ParsedArgs& args,
                                                 storage::TsFileReader& reader) {
     std::vector<SeriesStatRow> rows;
+    const std::string table_filter = storage::to_lower(args.table);
```
jt2594838: It is weird to call a single value a filter; better to use `target_table` or else.

jt2594838 (approved):

cpp/tools/commands/statistics.cc
Alchuang22-dev: Sure, `target_table_name` maybe a clearer choice.

PR #848, opened 2026-06-24
fix tag null segsegv.

ColinLeeo:
## Summary
  The Python SDK **segfaulted** when reading a table-model TsFile in which a device
  has a `NULL` tag value (e.g. `14_boundary_table_nullable_tag_values_two_fields.tsfile`),
  and even with the crash worked around, such devices read back as all-`NaN` or got
  silently dropped. This PR makes null-tagged devices read correctly end to end
  (C++ core → C wrapper → Python), and adds first-class `IS NULL` support.

  ## Root cause
  A null tag is stored as a `nullptr` segment in the device id.

  1. **Crash (SIGSEGV).** Every C++ tag filter only checked the column-index bound,
     not for `nullptr`, then did `*segments[i]` — dereferencing a null pointer.
     (`lldb` pinned it to `storage::TagEq::satisfyRow`.)
  2. **Silent wrong data.** The C wrapper exposed a null segment to Python as the
     literal string `"null"`, while the query path kept the real `nullptr`. So the
     auto-generated `TagEq("null")` could never match the real-null device, and the
     metadata map (keyed by the device-name string, which renders null as `"null"`)
     collided a real null tag with the literal string `"null"` and dropped a device.
  3. **Lossy logical paths.** The dataset layer flattened a device to
     `table.tag.field` and *dropped* null tags, losing their position, so devices
     with nulls in different positions (`a.<null>.b.c` vs `a.b.<null>.c`) collided
     and became unaddressable.

  ## API / behavior changes
  - `list_timeseries()` now returns `List[SeriesPath]` (still `str`, backward
    compatible). Null tags keep their position via `\N`, e.g.
    `eval_nullable_tags.\N.sensor_A.temperature`.
  - `loc[time, series]` accepts a path string (`\N` = null) or a `SeriesPath`,
    which can be built three equivalent ways:

        series = df.list_timeseries()
        df.loc[:, series]                                          # round-trips losslessly
        df.loc[:, "eval_nullable_tags.\\N.sensor_A.temperature"]  # string + \N
        df.loc[:, SeriesPath(["eval_nullable_tags", None, "sensor_A", "temperature"])]
        #         flat [table, *tags, field];  None = null tag

  - Over-specified paths (too many tags) now fail with `Series not found` instead
    of being silently truncated into a wrong match.

  ## Tests
  - C++ gtest: null-segment safety for every comparison filter, plus
    `TagIsNull` / `TagIsNotNull` (including absent trailing segments).
  - Python: null-position preservation, real-null vs the string `"null"`,
    `SeriesPath` construction forms and escaping round-trip, and rejection of
    over-specified paths. Full Python suite: **149 passing**.

### DIFF
--- a/cpp/src/reader/filter/tag_filter.cc
+++ b/cpp/src/reader/filter/tag_filter.cc
@@ -44,7 +44,8 @@ TagEq::TagEq(int col_idx, std::string tag_value)
     : TagFilter(col_idx, std::move(tag_value)) {}
 
 bool TagEq::satisfyRow(std::vector<std::string*> segments) const {
-    if (col_idx_ >= segments.size()) return false;
+    if (col_idx_ >= segments.size() || segments[col_idx_] == nullptr)
+        return false;
     return *segments[col_idx_] == value_;
 }
 
@@ -53,7 +54,8 @@ TagNeq::TagNeq(int col_idx, std::string tag_value)
     : TagFilter(col_idx, std::move(tag_value)) {}
 
 bool TagNeq::satisfyRow(std::vector<std::string*> segments) const {
-    if (col_idx_ >= segments.size()) return false;
+    if (col_idx_ >= segments.size() || segments[col_idx_] == nullptr)
+        return false;
     return *segments[col_idx_] != value_;
 }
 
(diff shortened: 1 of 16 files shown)


cpp/src/reader/filter/tag_filter.cc
```diff
@@ -254,6 +282,18 @@ Filter* TagFilterBuilder::not_reg_exp(const std::string& columnName,
     return new TagNotRegExp(idx, value);
 }
 
+Filter* TagFilterBuilder::is_null(const std::string& columnName) {
+    auto idx = get_id_column_index(columnName);
```
jt2594838: id -> tag

hongzhi-gao (approved):
Please update the Python API docs: get_timeseries_metadata() now keys by segments tuple (with None for null tags)

jt2594838 (approved):

PR #862, opened 2026-07-09
[Python][C++] Fix tree-model TsFileDataFrame returning empty data on reused readers and uppercase names

Young-Leo:
### Summary

When reading tree-model TsFiles through `TsFileDataFrame`, series reads could
return empty arrays in two situations. This PR fixes both, which together were
responsible for "read one series OK, then read another series returns empty" on
a reused reader.

Both fixes go on top of the existing tree-model support (#816).

---

### Fix 1 — C++: make `StringArrayDeviceID::split_table_name()` idempotent

**Problem**
Device IDs are cached and reused across queries. `split_table_name()` /
`init_prefix_segments()` appended to `prefix_segments_` every time it was
called, so on a reused reader the prefix segments accumulated across successive
queries. The stale/duplicated prefix segments corrupted the per-device path
columns, so a second series read on the same reader could match nothing and
return empty (and it also leaked the previously allocated segment pointers).

**Change**
`init_prefix_segments()` now clears and frees the existing `prefix_segments_`
before rebuilding, making `split_table_name()` idempotent. Repeated calls
produce a stable, correct segment vector.

Files: `cpp/src/common/device_id.cc`, test `cpp/test/common/device_id_test.cc`
(`DeviceIdTest.SplitTableNameIsIdempotent`).

---

### Fix 2 — Python: read series with uppercase measurement / path names

**Problem**
Tree TsFiles have no table schema, so the Python dataset layer reads them by
synthesizing a virtual table via `query_table_on_tree`. Because the table model
is case-insensitive, the native layer normalizes the synthesized column names
and path segments to lower case (e.g. `Temperature` -> `temperature`), while
the dataset catalog keeps the original casing from timeseries metadata. The two
tree read paths then looked the field column and device path up
case-sensitively, so an uppercase measurement or path segment failed the lookup
and returned an empty array:

```python
with TsFileDataFrame(path) as tsdf:
    tsdf["root.ln.wf01.wt01.Temperature"][:]
    # -> array([], dtype=float64)   # expected [0.5, 1.5, 2.5, 3.5, 4.5]
```

__Change__ Match field column names and device path segments case-insensitively in `_read_series_by_row_tree` and `_read_arrow_tree`, consistent with the case-insensitive table model. The original casing is still surfaced to users; only the internal lookup against the lower-cased result set is normalized.

Files: `python/tsfile/dataset/reader.py`, regression test in `python/tests/test_tsfile_dataset.py`.

---

### Tests

- New C++ regression test: `DeviceIdTest.SplitTableNameIsIdempotent`.
- New Python regression tests covering reused-reader reads and uppercase measurement/path names.
- No public API change; logical series names keep their original casing.

### DIFF
--- a/cpp/src/common/device_id.cc
+++ b/cpp/src/common/device_id.cc
@@ -103,6 +103,13 @@ std::string StringArrayDeviceID::get_device_name() const {
 }
 
 void StringArrayDeviceID::init_prefix_segments() {
+    // Idempotent: device IDs are cached and reused across queries, so clear
+    // previous prefixes before rebuilding to avoid accumulation and leaks.
+    for (const auto& prefix_segment : prefix_segments_) {
+        delete prefix_segment;
+    }
+    prefix_segments_.clear();
+
 #ifdef ENABLE_ANTLR4
     auto splits = storage::PathNodesGenerator::invokeParser(*segments_[0]);
 #else
@@ -130,6 +137,7 @@ int StringArrayDeviceID::serialize(common::ByteStream& write_stream) {
 
 int StringArrayDeviceID::deserialize(common::ByteStream& read_stream) {
     int ret = common::E_OK;
+
     uint32_t num_segments;
     if (RET_FAIL(common::SerializationUtil::read_var_uint(num_segments,
                                                           read_stream))) {
--- a/cpp/src/common/schema.h
+++ b/cpp/src/common/schema.h
@@ -209,7 +209,6 @@ class TableSchema {
             to_lowercase_inplace(table_name_);
         }
         for (const common::ColumnSchema& column_schema : column_schemas) {
-
             column_schemas_.emplace_back(std::make_shared<MeasurementSchema>(
                 column_schema.get_column_name(),
                 column_schema.get_data_type()));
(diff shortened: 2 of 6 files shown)


cpp/src/common/device_id.cc
```diff
@@ -103,6 +103,13 @@ std::string StringArrayDeviceID::get_device_name() const {
 }
 
 void StringArrayDeviceID::init_prefix_segments() {
+    // Idempotent: device IDs are cached and reused across queries, so clear
+    // previous prefixes before rebuilding to avoid accumulation and leaks.
+    for (const auto& prefix_segment : prefix_segments_) {
+        delete prefix_segment;
+    }
+    prefix_segments_.clear();
+
```
ColinLeeo, line 112: It's not a good idea but useful...

ColinLeeo:
In the tree model, both device paths and measurements are case-sensitive, so we should not simply normalize them to lowercase. Otherwise, case-distinct paths/measurements may be overwritten or mixed. The Python layer should preserve the original casing and resolve values by exact column position. On the C++ side, the tree query path does not seem to lowercase these names directly, but ResultSet name-based lookup is case-insensitive, so there may still be a similar collision risk that we should verify/fix as well.

ColinLeeo (approved):

PR #872, opened 2026-07-21
Refactor data type handling with type interfaces

jt2594838:
## Summary

- introduce type interfaces and services for data-type-specific operations
- replace direct `TSDataType` switch logic across read, write, tools, examples, and tests
- move casting, sizing, numeric, and comparable behavior into type or enum-specific implementations
- preserve unsupported-type handling and route new user-facing messages through i18n resources

## Testing

- `./mvnw test -P with-java -pl java/tsfile -am -Dtest=TypeCastTest -Dsurefire.failIfNoSpecifiedTests=false`
- Spotless and Checkstyle validation


java/tsfile/src/main/java/org/apache/tsfile/read/common/DescReadWriteBatchData.java
```diff
@@ -49,43 +49,7 @@ public DescReadWriteBatchData(TSDataType dataType) {
...
-        binaryRet.add(new Binary[capacity]);
-        break;
-      case VECTOR:
-        vectorRet = new LinkedList<>();
-        vectorRet.add(new TsPrimitiveType[capacity][]);
-        break;
-      default:
-        throw new UnSupportedDataTypeException(String.valueOf(dataType));
-    }
+    Type.fromTsDataType(dataType).init(this);
```
hongzhi-gao, line 52: DescReadWriteBatchData initializes value lists as ArrayLists but casts them to LinkedList when expanding. This causes a ClassCastException on the 1,025th value.

hongzhi-gao (approved):

PR #873, opened 2026-07-23
Optimize Gorilla batch floating-point decoding

ColinLeeo:
## Motivation

The Gorilla batch decoder previously refilled its bit cache one byte at a time and decoded floating-point values through small temporary integer buffers. Profiling showed that most CPU time was spent in control-bit parsing, variable-width bit reads, and repeated batch bookkeeping.

This PR reduces that overhead while keeping the Gorilla encoding format unchanged.

## Modifications

- Replace the one-byte bit cache with a 64-bit bit reservoir.
  - Refill up to eight encoded bytes at once.
  - Preserve prefetched bits across batch, scalar, and skip operations.
- Add a fast path for variable-width bit reads.
  - Reads contained in the current reservoir avoid the generic loop.
  - Cross-reservoir reads require at most one additional refill.
- Decode floating-point batches directly into `float` and `double` output buffers.
  - Remove the temporary 129-value integer buffers.
  - Remove the associated copy/conversion pass.
- Use unsigned bit accumulation to avoid signed-shift undefined behavior.
- Validate truncated input and invalid Gorilla metadata.
- Fix decoder state handling when switching from scalar decoding to batch decoding.
- Preserve support for both wrapped and non-wrapped `ByteStream` inputs.

## Performance

Measured on ARM64 macOS using a Release build with `-O3` and LTO. Results are the median of three runs over 500,000 smooth floating-point values.

| Type | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| Float batch decode | 3.04 ns/value | 1.73 ns/value | 1.76x faster |
| Double batch decode | 5.43 ns/value | 2.12 ns/value | 2.57x faster |

Profiling also showed that the temporary-buffer copy was eliminated and the proportion of samples in variable-width bit reading decreased significantly.

## Compatibility

- The encoded Gorilla data format is unchanged.
- Existing encoded files remain compatible.
- Integer, long, float, and double Gorilla decoders share the optimized bit reader.
- Scalar, batch, and skip APIs can be interleaved safely.

## Tests

Added coverage for:

- Decoding one value per batch call.
- Interleaving scalar, batch, and skip operations.
- Full-width 64-bit XOR values.
- Non-wrapped floating-point batch input.
- Truncated input handling.

Validation results:

- 23 targeted Gorilla tests passed.
- 716 complete C++ tests passed.
- 10 pre-existing tests remain disabled.

### DIFF
--- a/cpp/src/encoding/gorilla_decoder.h
+++ b/cpp/src/encoding/gorilla_decoder.h
@@ -283,6 +283,7 @@ class GorillaDecoder : public Decoder {
         first_value_was_read_ = false;
         has_next_ = false;
         buffer_ = 0;
+        read_status_ = common::E_OK;
     }
 
     FORCE_INLINE bool has_next() { return has_next_; }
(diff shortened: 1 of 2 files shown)


jt2594838 (approved):
Is this referable to the Java edition?

ColinLeeo:
> Is this referable to the Java edition?

Yes, the core idea can be referenced for the Java implementation. 
Java’s GorillaDecoderV2 also uses an 8-bit buffer and refills it byte by byte from ByteBuffer, so a wider bit reservoir may reduce refill and branching overhead. 

However, this PR is C++-specific because it also relies on the C++ batch-decoding API and direct writes into float/double output arrays. Java currently decodes values individually into page/column builders, so it would require a separate implementation and Java-specific benchmarks before we can confirm the benefit.

PR #882, opened 2026-07-29
Support TsFile properties and optional table point counting

jt2594838:
## Summary

- add `TsFileProperties` read/write APIs to `TsFileIOWriter`, `TsFileSequenceReader`, `ITsFileWriter`, and `ITsFileReader`
- add an optional `TsFileIOWriter` mode that records non-null FIELD point counts per table in TsFile properties
- add a repair tool that detects and backfills missing table point-count properties in existing complete TsFiles
- package Linux and Windows launch scripts and document the tool in the Java tools README
- add API, repair-tool, and performance-test coverage

## Point-count semantics

- properties use the `tablePointCount.<tableName>` key format
- timestamps, TAG values, and null FIELD values are excluded
- collection is disabled by default

## Table point-count repair tool

- scans table-model metadata and calculates missing point counts without enumerating devices
- preserves existing data and unrelated TsFile properties
- leaves files unchanged when all table point-count properties are already present
- rejects missing or incomplete TsFiles
- writes and flushes a temporary file before replacing the source
- uses reflink copy-on-write on supported Linux and macOS file systems, with a full prefix-copy fallback
- provides `tsfile-table-point-count.sh` and `tsfile-table-point-count.bat`

## Verification

- `mvn -P with-java -pl java/tsfile -am test-compile -DskipTests`
- repair-tool tests cover multi-table counting, null FIELD exclusion, property preservation, file completeness, and idempotent updates
- disabled mode: PID 7620, 36,350 bytes, 67.309 ms
- enabled mode: PID 25216, 36,410 bytes, 74.772 ms
- file-size overhead: 60 bytes (0.165%)

Each performance mode performed an independent warm-up in its own Surefire JVM.

### DIFF
--- a/java/tools/README.md
+++ b/java/tools/README.md
@@ -228,3 +228,27 @@ arrow2tsfile.bat --source .\data\arrow --target .\output --fail_dir .\failed
 - Multiple batches: `{source_basename}_1.tsfile`, `{source_basename}_2.tsfile`, ...
 - Table name and output filename are independent — table name comes from schema or `--table_name`, filename comes from source file.
 
+## Table Point Count Tool
+
+`tools\tsfile-table-point-count.bat` checks whether a complete TsFile contains table-level
+point-count properties and backfills them when they are missing. For each table, the point count is
+the total number of non-null values in its FIELD columns; TAG and time columns are not counted.
+
+Set `JAVA_HOME`, then run the tool with exactly one TsFile path:
+
+```bat
+tools\tsfile-table-point-count.bat C:\data\example.tsfile
+```
+
+The tool prints one of the following statuses:
+
+| Status | Description |
+|--------|-------------|
+| `UPDATED` | Missing point-count properties were calculated and written to the file. |
+| `ALREADY_PRESENT` | Every table already had a valid point-count property; the file was not modified. |
+| `NO_TABLE` | The file contains no table schema, for example a tree-model-only TsFile; the file was not modified. |
+
+The tool modifies the input file in place only for `UPDATED`. It writes a complete temporary copy
+next to the source file and replaces the source after the rewritten metadata has been flushed. Make
+sure the directory is writable, leave enough free space for a temporary copy, and back up important
+files before running the tool. Missing files and incomplete TsFiles are rejected.
--- a/java/tools/README.md
+++ b/java/tools/README.md
@@ -230,15 +230,17 @@ arrow2tsfile.bat --source .\data\arrow --target .\output --fail_dir .\failed
 
 ## Table Point Count Tool
 
-`tools\tsfile-table-point-count.bat` checks whether a complete TsFile contains table-level
+`tools/tsfile-table-point-count.sh` (Linux/macOS) and `tools\tsfile-table-point-count.bat` (Windows) check whether a complete TsFile contains table-level
 point-count properties and backfills them when they are missing. For each table, the point count is
 the total number of non-null values in its FIELD columns; TAG and time columns are not counted.
 
 Set `JAVA_HOME`, then run the tool with exactly one TsFile path:
 
-```bat
-tools\tsfile-table-point-count.bat C:\data\example.tsfile
-```
+    # Linux/macOS
+    tools/tsfile-table-point-count.sh /data/example.tsfile
+
+    :: Windows
+    tools\tsfile-table-point-count.bat C:\data\example.tsfile
 
 The tool prints one of the following statuses:
 
--- a/java/tools/src/assembly/tools.xml
+++ b/java/tools/src/assembly/tools.xml
@@ -69,5 +69,14 @@
             <source>${maven.multiModuleProjectDirectory}/java/tools/src/assembly/resources/tools/arrow2tsfile.bat</source>
             <destName>tools/arrow2tsfile.bat</destName>
         </file>
+        <file>
+            <source>${maven.multiModuleProjectDirectory}/java/tools/src/assembly/resources/tools/tsfile-table-point-count.sh</source>
+            <destName>tools/tsfile-table-point-count.sh</destName>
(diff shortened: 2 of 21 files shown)


java/tsfile/src/main/java/org/apache/tsfile/read/TsFileSequenceReader.java
```diff
@@ -607,6 +607,11 @@ public Map<String, TableSchema> getTableSchemaMap() throws IOException {
     return getTableSchemaMap(null);
   }
 
+  /** Get the properties stored in the TsFile metadata. */
+  public Map<String, String> getTsFileProperties() throws IOException {
```
jt2594838: This exposes TsFile properties through the low-level reader API so callers can retrieve file-level metadata without depending on internal metadata objects. The corresponding V4 reader/writer interfaces expose the same capability across implementations while preserving existing behavior for callers that do not use properties.

java/tsfile/src/main/java/org/apache/tsfile/write/writer/TsFileIOWriter.java
```diff
@@ -421,6 +445,15 @@ public void writeChunk(Chunk chunk) throws IOException {
   public void endCurrentChunk() {
     this.currentChunkMetadataSize += currentChunkMetadata.getRetainedSizeInBytes();
     chunkMetadataCount++;
+    if (recordTablePointCount
```
jt2594838, line 448: Point counting is opt-in to avoid changing existing write-path cost or file output by default. It aggregates only non-null FIELD values into `tablePointCount.<tableName>` properties; the mask check excludes timestamp chunks, while table-model device detection excludes tree-model data.

java/tsfile/src/main/java/org/apache/tsfile/utils/TsFileTablePointCountTool.java
```diff
@@ -0,0 +1,318 @@
...
+          && hasAllTablePointCountProperties(reader.getTsFileProperties(), tableSchemas.keySet());
+    }
+  }
+
+  /**
+   * Adds table point-count properties when they are missing. Existing complete properties are left
+   * untouched. The source file is replaced only after a complete rewritten copy has been forced to
+   * disk.
+   */
+  public static UpdateStatus updateTablePointCountIfMissing(File file) throws IOException {
```
jt2594838, line 92: This entry point backfills missing counts for complete table-model TsFiles by scanning timeseries metadata instead of enumerating every device. It preserves unrelated properties, skips files that already have valid counts, and replaces the source only after a temporary rewrite has been flushed; reflink cloning is used where available with a prefix-copy fallback.

java/tools/src/assembly/tools.xml
```diff
@@ -69,5 +69,14 @@
             <source>${maven.multiModuleProjectDirectory}/java/tools/src/assembly/resources/tools/arrow2tsfile.bat</source>
             <destName>tools/arrow2tsfile.bat</destName>
         </file>
+        <file>
+            <source>${maven.multiModuleProjectDirectory}/java/tools/src/assembly/resources/tools/tsfile-table-point-count.sh</source>
```
jt2594838, line 73: Packaging both launchers makes the repair utility directly available from the assembled tools distribution. The shell entry is explicitly marked executable, and the Windows launcher is placed beside it using the same tool name.

java/tools/README.md
```diff
@@ -228,3 +228,27 @@ arrow2tsfile.bat --source .\data\arrow --target .\output --fail_dir .\failed
 - Multiple batches: `{source_basename}_1.tsfile`, `{source_basename}_2.tsfile`, ...
 - Table name and output filename are independent — table name comes from schema or `--table_name`, filename comes from source file.
 
+## Table Point Count Tool
```
jt2594838, line 231: This section documents the tool point-count semantics, invocation, result statuses, in-place replacement behavior, and disk-space and backup considerations. The matching Chinese README and message bundles keep the user experience consistent with the project localization convention.

java/tsfile/src/test/java/org/apache/tsfile/utils/TsFileTablePointCountToolTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+public class TsFileTablePointCountToolTest {
+
+  /**
+   * Creates a two-table file without point-count properties. The repair tool must traverse the
+   * metadata index tree and count only non-null FIELD values, preserve unrelated properties and
+   * file data, and leave the file byte-for-byte unchanged when run again.
+   */
+  @Test
+  public void backfillMissingTablePointCountProperties() throws Exception {
```
jt2594838, line 51: This test covers multi-table counting, null FIELD exclusion, preservation of unrelated properties and file data, and idempotent re-execution. The related API and writer tests cover property round trips and opt-in collection, while the performance test isolates enabled and disabled modes in separate JVMs to avoid warm-up bias.

java/tsfile/src/main/java/org/apache/tsfile/utils/TsFileTablePointCountTool.java
```diff
@@ -0,0 +1,318 @@
...
+      }
+      tablePointCounts = scanTablePointCounts(reader, tableSchemas);
+      fileMetadataPosition = reader.getFileMetadataPos();
+    }
+
+    tablePointCounts.forEach(
+        (tableName, pointCount) ->
+            metadata.addProperty(
+                TsFileIOWriter.TABLE_POINT_COUNT_PROPERTY_PREFIX + tableName,
+                Long.toString(pointCount)));
```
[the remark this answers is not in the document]
jt2594838: This is intentionally overwritten by design. The user should not specify this or use our key.

PR #909, opened 2026-08-18
fix(cpp): never seal empty chunks and fix dangling ref in parallel tablet write

### LINKED ISSUE #908: fix(cpp): never seal empty chunks and fix dangling ref in parallel tablet write
### Bug 1: registered-but-empty measurements are sealed as EMPTY chunks, making Java readers refuse the file

**Environment:** TsFileCpp 2.3.2.dev (cpp tree-model writer)

When a tree-model `TsFileWriter` registers a measurement but receives no data for it during a window (e.g. event-driven parameters that fire sporadically, or a wide schema where only a subset of columns receive values), `flush_chunk_group()` / `flush_chunk_group_encoded()` seal that column as an **empty chunk**:

```
[Chunk] of <device>.<measurement>, startTime: 9223372036854775807 endTime: -9223372036854775808 count: 0
[Chunk Header] marker=5, dataSize=0, size=78
```

Java readers treat this as a crashed file — `TsFileSequenceReader` self-check fails (`self-check cannot proceed at position ... because: 0`) and `TsFileSketchTool` throws `IOException: ... because the file has crashed` plus an NPE while printing the empty chunk. The whole file becomes unreadable by the Java toolchain, even though most chunks contain valid data.

**Cause:** the non-aligned branch of both flush functions only checks `m_schema->chunk_writer_ != nullptr` and lacks the `hasData()` guard that the aligned branch already has:

```cpp
if (!chunk_group->is_aligned_ && m_schema->chunk_writer_ != nullptr) {
    FLUSH_CHUNK(chunk_writer, ...);   // seals count=0 chunk when empty
}
```

**Expected behavior:** an empty column should produce no chunk (mirroring the aligned branch and Java IoTDB behavior).

---

### Bug 2: dangling reference capture in parallel aligned tablet write

In `write_table()` (aligned parallel path), the lambdas submitted to the thread pool capture the **loop variables by reference**:

```cpp
for (auto& ctx : device_ctxs) {
    futures.push_back(common::g_thread_pool_->submit(
        [&write_time_segments, &ctx]() { ... }));   // captures loop var by reference
    for (auto& vt : ctx.value_tasks) {
        futures.push_back(common::g_thread_pool_->submit(
            [&write_value_segments, &vt, &ctx]() { ... }));
    }
}
```

> [!NOTE]
> **Correction (post-review):** as discussed in the PR #909 review, this is *not* an actual dangling-reference bug in the current code — `device_ctxs` is fully populated before any task is submitted, and every future is joined via `f.get()` before the vector goes out of scope, so the by-reference captures are lifetime-safe today. The capture-by-value change in PR #909 is a defensive cleanup that removes the implicit dependency on the submission loop's scope. **Bug 1 (empty chunk sealing) stands on its own as reported.**

`submit` enqueues the task for asynchronous execution on pool threads. The by-reference captures of the loop variables rely on `device_ctxs` / `ctx.value_tasks` outliving all joined futures — an invariant that holds in the current code shape but is easy to break under refactoring (e.g. moving the `get()` loop out of the submission scope).

**Expected behavior:** capture the per-iteration addresses by value (`ctx_ptr`/`vt_ptr`) to make the per-task address explicit (lifetime-equivalent to the old captures; task-owned lifetime would require copying the inputs).

kkzi:
Fixes #908

### 1. Fix: never seal empty chunks (non-aligned flush path)

`flush_chunk_group()` / `flush_chunk_group_encoded()` now skip registered-but-empty measurement columns via the existing `ChunkWriter::hasData()` check — mirroring what the aligned branch already does.

**Before:** a measurement that received no data in a window was sealed as an EMPTY chunk (`count=0, dataSize=0`). Java readers (`TsFileSequenceReader` self-check / `TsFileSketchTool`) treat the whole file as crashed and refuse to load it, even though the other chunks are valid.

**After:** empty columns produce no chunk at all.

### 2. Cleanup: capture per-iteration state by value in parallel aligned tablet write

In `write_table()`'s aligned parallel path, the tasks submitted to the thread pool captured the loop variables (`ctx`, `vt`) **by reference**. As discussed in the review, those references were in fact lifetime-safe in the current code shape: `device_ctxs` is fully populated before any task is submitted and every future is joined via `f.get()` before the vector goes out of scope — so this was not an actual dangling-reference bug. Each by-reference capture also referred to its own vector element (not a shared loop slot), so the capture-by-pointer change (`ctx_ptr` / `vt_ptr`) is equivalent in lifetime and merely makes the per-task address explicit. Truly task-owned lifetime would require copying the task inputs; this is a readability/robustness cleanup, not a lifetime fix.

## Verification

- Both changes compile cleanly with the existing MSVC/Ninja build (only `tsfile_writer.cc` changed, no build config touched).
- Change 1 validated end-to-end: a reproducer that registers 3 measurements but writes to only 2 now produces a file with 2 chunks (no `count=0` chunk), and `TsFileSketchTool` / `print-tsfile.bat` read it completely (`END of TsFile`, `TsFile Sketch End`, no errors).
- 17 regression tests added (see 3a4c59f and cf742de): 10 for the empty-chunk fix (tablet/record write paths, multi-window flush, memory auto-flush, partially-null and TEXT counter-checks, mixed aligned/non-aligned) and 7 for the parallel aligned write path (multi-device/multi-column round-trips crossing page boundaries, tiny page size, thread-count boundaries). Verified in both directions: with the `hasData()` guards reverted the empty-chunk tests fail; with the changes in place all 17 pass along with the full TsFile_Test suite (767 tests).

### DIFF
--- a/cpp/src/writer/tsfile_writer.cc
+++ b/cpp/src/writer/tsfile_writer.cc
@@ -1351,11 +1351,12 @@ int TsFileWriter::write_table(Tablet& tablet) {
             common::g_thread_pool_ != nullptr) {
             std::vector<std::future<int>> futures;
             for (auto& ctx : device_ctxs) {
-                // Capture the per-iteration state by pointer value. The
-                // references stay valid today (device_ctxs outlives all
-                // future.get() calls below), but the by-value form doesn't
-                // rely on that lifetime coincidence and stays safe if the
-                // get() loop is ever moved out of this scope.
+                // Capture the per-iteration state by pointer value. This
+                // is equivalent in lifetime to the old by-reference
+                // captures (each referred to its own vector element, and
+                // device_ctxs outlives all future.get() calls below) — it
+                // only makes the per-task address explicit. Truly task-
+                // owned lifetime would require copying the task inputs.
                 auto* ctx_ptr = &ctx;
                 futures.push_back(common::g_thread_pool_->submit(
                     [&write_time_segments, ctx_ptr]() {
--- a/cpp/src/writer/tsfile_writer.cc
+++ b/cpp/src/writer/tsfile_writer.cc
@@ -1351,10 +1351,11 @@ int TsFileWriter::write_table(Tablet& tablet) {
             common::g_thread_pool_ != nullptr) {
             std::vector<std::future<int>> futures;
             for (auto& ctx : device_ctxs) {
-                // Capture pointers by value: the submitted tasks run on pool
-                // threads asynchronously (after this loop returns), so
-                // capturing the loop variables by reference would dangle
-                // (all tasks would read the same/out-of-scope ctx).
+                // Capture the per-iteration state by pointer value. The
+                // references stay valid today (device_ctxs outlives all
+                // future.get() calls below), but the by-value form doesn't
+                // rely on that lifetime coincidence and stays safe if the
+                // get() loop is ever moved out of this scope.
                 auto* ctx_ptr = &ctx;
                 futures.push_back(common::g_thread_pool_->submit(
                     [&write_time_segments, ctx_ptr]() {
--- a/cpp/src/writer/tsfile_writer.cc
+++ b/cpp/src/writer/tsfile_writer.cc
@@ -1358,17 +1358,17 @@ int TsFileWriter::write_table(Tablet& tablet) {
                 auto* ctx_ptr = &ctx;
                 futures.push_back(common::g_thread_pool_->submit(
                     [&write_time_segments, ctx_ptr]() {
-                        return write_time_segments(ctx_ptr->tcw,
-                                                   ctx_ptr->segments,
-                                                   ctx_ptr->initial_page_points);
+                        return write_time_segments(
+                            ctx_ptr->tcw, ctx_ptr->segments,
+                            ctx_ptr->initial_page_points);
                     }));
                 for (auto& vt : ctx.value_tasks) {
                     auto* vt_ptr = &vt;
                     futures.push_back(common::g_thread_pool_->submit(
                         [&write_value_segments, vt_ptr, ctx_ptr]() {
                             return write_value_segments(
-                                vt_ptr->vcw, vt_ptr->col_idx,
-                                ctx_ptr->segments, ctx_ptr->initial_page_points);
+                                vt_ptr->vcw, vt_ptr->col_idx, ctx_ptr->segments,
+                                ctx_ptr->initial_page_points);
                         }));
                 }
             }
--- a/cpp/src/writer/tsfile_writer.cc
+++ b/cpp/src/writer/tsfile_writer.cc
@@ -1351,17 +1351,24 @@ int TsFileWriter::write_table(Tablet& tablet) {
             common::g_thread_pool_ != nullptr) {
             std::vector<std::future<int>> futures;
             for (auto& ctx : device_ctxs) {
+                // Capture pointers by value: the submitted tasks run on pool
+                // threads asynchronously (after this loop returns), so
+                // capturing the loop variables by reference would dangle
+                // (all tasks would read the same/out-of-scope ctx).
+                auto* ctx_ptr = &ctx;
                 futures.push_back(common::g_thread_pool_->submit(
-                    [&write_time_segments, &ctx]() {
-                        return write_time_segments(ctx.tcw, ctx.segments,
-                                                   ctx.initial_page_points);
+                    [&write_time_segments, ctx_ptr]() {
+                        return write_time_segments(ctx_ptr->tcw,
+                                                   ctx_ptr->segments,
+                                                   ctx_ptr->initial_page_points);
                     }));
                 for (auto& vt : ctx.value_tasks) {
+                    auto* vt_ptr = &vt;
                     futures.push_back(common::g_thread_pool_->submit(
-                        [&write_value_segments, &vt, &ctx]() {
+                        [&write_value_segments, vt_ptr, ctx_ptr]() {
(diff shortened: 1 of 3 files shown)


kkzi:
Pushed `6363b02` to fix the spotless clang-format violations in tsfile_writer.cc.

CI runs for the new commit are waiting for approval (`action_required`) — could you approve the workflows to re-run them?

ColinLeeo:
Hi @kzzi, thanks for creating the PR. 
After making the code changes, I think we should also add some regression tests to verify that the fix works as expected.

kkzi:
Hi @ColinLeeo, thanks for the review!

Regression tests have been added in 3a4c59f and cf742de (17 tests total, all passing along with the full TsFile_Test suite — 767 tests):

**Empty-chunk sealing (fix 1)** — `cpp/test/writer/tsfile_writer_empty_chunk_regression_test.cc`:
- tablet write path, record write path, multi-window flush, empty-window-then-write, sibling devices, and mixed aligned/non-aligned devices; each asserts via `TsFileReader::get_timeseries_metadata()` that an unwritten measurement is absent from the file while surviving columns carry real statistics (count/start/end)
- memory-threshold auto-flush entry (`check_memory_size_and_may_flush_chunks()`), a second path into `flush_chunk_group()` that explicit-flush tests never drive
- counter-check: a partially-null column is still sealed (the `hasData()` guard must only skip fully empty columns)
- TEXT column variant (distinct string write path)

**Parallel aligned tablet write (fix 2)** — `cpp/test/writer/table_view/tsfile_parallel_tablet_write_regression_test.cc`:
- multi-device × multi-column × rows crossing page boundaries, single-device many-columns, 64-device vector-reallocation, sequential batches, flush-between-batches, tiny page size (8 points), and 1-/8-thread pool configs — all round-trip and verify every cell (row completeness, per-cell values, tag correctness)

Verified both directions: with the `hasData()` guards reverted, the empty-chunk tests fail (the empty column appears in the file's metadata); with the fixes in place everything passes.

One note for completeness: the parallel-write tests are end-to-end coverage of the thread-pool path rather than a deterministic reproducer of the dangling capture — `device_ctxs` is fully built before the submission loop and all `future.get()` calls happen within its lifetime, so the old by-reference captures happened to be lifetime-safe in the current code shape. The fix removes the hazard for future refactors (e.g. moving `get()` out of scope).

kkzi:
CI runs for the new test commits (`3a4c59f`, `cf742de`) are waiting for approval (`action_required`) — could a maintainer approve the workflows so they re-run?

ColinLeeo:
Thanks for the fix, and for the thorough test coverage — the empty-chunk fix (fix #1) looks solid to me.

One thing I'd like to flag on fix #2 (the "dangling reference" in the parallel aligned tablet write path): I don't think this bug actually existed in the original code. As Copilot's review also notes, `ctx`/`vt` are references bound to vector elements — and those vectors (`device_ctxs`, `value_tasks`) are fully populated before any task is submitted, and every future is joined via `f.get()` before `device_ctxs` goes out of scope. So the old `[&ctx]`/`[&vt]` captures shouldn't have dangled. The capture-by-pointer change itself is harmless, so I don't think it needs to be reverted — but would it be possible to reword the PR description and the new test comments so they don't describe it as fixing a dangling-reference bug? I'd rather we not leave that framing in the history if it isn't quite accurate.

Also, echoing Copilot's other comment: `TinyPageBoundaryRoundTrip` currently only validates rows that the query actually returns, without asserting that every expected row was seen — so a dropped row on the page-boundary path wouldn't necessarily fail the test. Would you mind adding a completeness check there (e.g. asserting every entry in `seen` is set, or comparing a row count), similar to what `VerifyTableRoundTrip` / `SequentialBatchesRoundTrip` already do?

Happy to help with either of these if useful — otherwise this looks good to merge once addressed.

kkzi:
Thanks @ColinLeeo — you're right on both counts, and I appreciate the careful review.

**On fix #2:** agreed, the by-reference captures were lifetime-safe in the current code shape (`device_ctxs` is fully populated before submission and every future is joined before the vector goes out of scope), so "dangling reference bug" was the wrong framing — I had reproduced the failure scenario against an imagined refactored shape rather than the code as written. Apologies for the noise; issue #908's bug 2 section now carries a correction note, and I've reworded the PR description and the source/test comments (a9af789) to describe the change as a defensive cleanup instead.

**On `TinyPageBoundaryRoundTrip`:** good catch — it only checked the rows the query returned, so a dropped row would have passed silently. a9af789 adds a completeness assertion: after each batch's query loop, every entry in `seen` must be set, so any missing row fails the test (same guarantee as `VerifyTableRoundTrip` / `SequentialBatchesRoundTrip`).

All 17 tests still pass with the changes, along with the full TsFile_Test suite.

kkzi:
Also addressing Copilot's review comments directly (thanks both):

- Both inline comments (the lifetime point and the `TinyPageBoundaryRoundTrip` completeness gap) are handled — the completeness assertion went in with a9af789.
- On the lifetime point, Copilot is right that `ctx_ptr`/`vt_ptr` are lifetime-equivalent to the old by-reference captures (each referred to its own vector element, not a shared loop slot). My reworded comments still overclaimed that the by-value form "stays safe if the get() loop is ever moved out of this scope" — it wouldn't, and true task-owned lifetime would require copying the task inputs. f5a3847 corrects the source/test comments (and the PR description) to state the equivalence instead of implying a stronger guarantee. No code changes.

ColinLeeo:
LGTM, thanks for the updates.

ColinLeeo (approved):

PR #915, opened 2026-08-20
Update C++ tsfile-cli command surface and data workflows

SpriCoder:
### Summary
- `ls`: lists model/object rows for current files; example: `tsfile-cli ls -f csv data.tsfile`
- `schema`: reports `model,object,column,category,data_type,encoding,compression`; example: `tsfile-cli schema -t sensors -f csv data.tsfile`
- `meta`: reports file summary fields; example: `tsfile-cli meta data.tsfile`
- `stats`: reports FIELD statistics with counts, null counts, time range, values, and source fields; example: `tsfile-cli stats -t sensors -m temperature -f csv data.tsfile`
- `count`: reports per-column counts without synthetic summary rows; example: `tsfile-cli count -t sensors -f csv data.tsfile`
- `sketch`: prints the physical file sketch; example: `tsfile-cli sketch data.tsfile`
- `head`: reads the first rows with offset/limit/filter support; example: `tsfile-cli head -t sensors -m temperature -n 20 -f csv data.tsfile`
- `cat`: streams all matching rows with the same row filters; example: `tsfile-cli cat -t sensors --tag-filter site eq north -m temperature -f ndjson data.tsfile`
- `export`: exports one object or multiple objects with numbered outputs and a manifest; example: `tsfile-cli export -t sensors --type csv -o sensors.csv data.tsfile`
- `write`: imports strict CSV into a new table-model TsFile with explicit `--tag` / `--field` declarations; example: `printf 'time,site,temperature\n0,north,21.5\n1,north,21.7\n' | tsfile-cli write --table sensors --tag site STRING --field temperature DOUBLE --stdin -o sensors.tsfile`

### Tests
- C++ Spotless check
- C++ CLI contract and end-to-end tests
- Full C++ `TsFile_Test` binary
- Git whitespace check

### Appendix
- Feishu requirement document: https://apache-iotdb-project.feishu.cn/docx/Hbd3dA7lco78ofxAn5jc3talnkb


cpp/test/tools/cli_args_test.cc
```diff
@@ -138,9 +164,9 @@ TEST(ParseArgsTest, MissingFileIsAllowedAtParseTime) {
 }
```
ColinLeeo: This test should be removed.

cpp/test/tools/cli_args_test.cc
```diff
@@ -97,27 +126,24 @@ TEST(ParseArgsTest, TagFilterParsed) {
...
 
 TEST(ParseArgsTest, DuplicateTagFilterIsError) {
     auto p = tsfile_cli::parse_args({"cat", "--tag-filter", "id1", "eq",
-                                     "dev_a", "--tag-between", "id1", "a", "z",
+                                     "dev_a", "--tag-filter", "id1", "neq", "z",
                                      "data.tsfile"});
-    EXPECT_FALSE(p.error.empty());
+    EXPECT_TRUE(p.error.empty());
+    ASSERT_EQ(p.tag_filters.size(), 2u);
 }
```
ColinLeeo: Multiple tag filters are supported, so this test should be removed.

cpp/tools/commands/row_query.cc
```diff
@@ -35,16 +37,119 @@
...
+   private:
+    std::regex pattern_;
+};
+
 bool can_push_down_row_window(const ParsedArgs& args, long long offset,
                               long long limit) {
-    return !args.has_start && !args.has_end && offset <= INT_MAX &&
+    return !args.has_start && !args.has_end && offset == 0 &&
            (limit < 0 || limit <= INT_MAX);
 }
```
ColinLeeo, line 60: It seems that both the time range and the offset can be pushed down here.

cpp/tools/commands/statistics.cc
```diff
@@ -136,49 +174,142 @@ StatisticCells statistic_value_cells(storage::Statistic* st) {
...
+            }
+            reader.destroy_query_data_set(result);
+            if (next_ret != common::E_OK) {
+                err << "Error: failed to count rows for device '" << device_name
+                    << "': " << error_code_message(next_ret) << "\n";
+                return kExitFile;
             }
-            std::string measurement =
-                ts->get_measurement_name().to_std_string();
+        }
```
ColinLeeo, line 267: This approach does not seem appropriate.

cpp/tools/commands/statistics.cc
```diff
@@ -191,24 +322,24 @@ FileSummary collect_file_summary(const ParsedArgs& args,
...
+            ++s.series_count;
+            storage::Statistic* statistic = series->get_statistic();
+            if (statistic == nullptr || statistic->get_count() <= 0) {
+                continue;
+            }
+            min_start = std::min(min_start, statistic->start_time_);
+            max_end = std::max(max_end, statistic->get_end_time());
+            s.has_time_range = true;
         }
-        min_start = std::min(min_start, row.start_time);
```
ColinLeeo, line 209: Could we obtain all series statistics here and aggregate them into device-level data? That should be feasible for aligned devices, at least.

cpp/tools/format/input_format.cc
```diff
@@ -34,6 +34,15 @@ void strip_cr(std::string& s) {
...
+std::string lower_ascii(const std::string& s) {
+    std::string out;
+    out.reserve(s.size());
+    for (unsigned char c : s) {
+        out += static_cast<char>(c >= 'A' && c <= 'Z' ? c + ('a' - 'A') : c);
+    }
+    return out;
+}
+
 }  // namespace
```
ColinLeeo: Lowercasing every value this way may be unnecessarily expensive. Is there an existing API we can reuse for this?

cpp/tools/format/input_format.cc
```diff
@@ -117,6 +126,14 @@ bool parse_columns_spec(const std::string& spec, std::vector<ColumnDef>& out,
...
+            return false;
+        }
+        if (def.category == common::ColumnCategory::TAG &&
+            def.type != common::STRING) {
+            error = "TAG column '" + def.name + "' must use STRING";
+            return false;
+        }
         for (const ColumnDef& prev : out) {
             if (prev.name == def.name) {
                 error = "duplicate column name '" + def.name + "'";
```
ColinLeeo: Doesn't this contradict the design document?

cpp/tools/format/input_format.cc
```diff
@@ -128,6 +145,79 @@ bool parse_columns_spec(const std::string& spec, std::vector<ColumnDef>& out,
...
+        if ((b == 0xE0 && bytes[i + 1] < 0xA0) ||
+            (b == 0xED && bytes[i + 1] > 0x9F) ||
+            (b == 0xF0 && bytes[i + 1] < 0x90) ||
+            (b == 0xF4 && bytes[i + 1] > 0x8F)) {
+            return false;
+        }
+        i += count + 1;
+    }
+    return true;
+}
```
ColinLeeo, line 159: The cost of this validation also seems quite high.

cpp/tools/format/result_set_format.cc
```diff
@@ -107,10 +108,20 @@ int emit_result_set(storage::ResultSet* rs, OutputFormat fmt, bool no_header,
                 cells[i - 1] = cell_to_string(rs, i, types[i - 1]);
             }
```
ColinLeeo, line 109: Shouldn't the reader be able to apply the filter here?

cpp/test/tools/cli_args_test.cc
ColinLeeo: Removed this test as suggested. The current write command accepts only explicit --tag/--field declarations, and the code and positive tests no longer use the old tag-between form.

cpp/tools/commands/statistics.cc
ColinLeeo: This concern is valid. The current implementation performs a full-range query for each tree device to obtain the logical row count. That is accurate for the timestamp union of non-aligned devices, but it is expensive. I have not replaced it with an approximation based on per-series counts; aligned devices could use the time-index statistics directly, while non-aligned or incomplete-statistics cases would still need a scan.

cpp/test/tools/cli_args_test.cc
ColinLeeo: Removed this test. Multiple --tag-filter options are supported and are combined with --tag-match all/any; the valid multi-condition path remains covered by the tests.

cpp/tools/format/input_format.cc
ColinLeeo: I rechecked requirement D-080: write type parameters accept only the canonical uppercase names, so int64 should be rejected. The parser and tests now enforce case-sensitive canonical spelling; this comment refers to the previous case-insensitive behavior.

cpp/tools/commands/statistics.cc
ColinLeeo: Addressed. collect_file_summary now calls get_timeseries_metadata() directly and aggregates the series count and time range for the whole file. It no longer invokes collect_series_stats and therefore does not trigger a per-device logical-row scan. Exact logical row counts for non-aligned devices remain a separate stats/count concern.

cpp/tools/format/input_format.cc
ColinLeeo: Addressed. input_format.cc now reuses the existing storage::to_lower API, and the local lower_ascii implementation has been removed.

cpp/tools/format/input_format.cc
ColinLeeo: I checked this validation. is_valid_utf8 performs a linear byte scan once per logical record and once for the header; it is not called per cell or recursively. There is no existing UTF-8 validation API in the project. validate_identifier performs separate BOM/control-character checks, but the total cost remains linear. We cannot remove this validation because requirement D-088 requires invalid UTF-8 to be rejected.

cpp/tools/commands/row_query.cc
ColinLeeo: I checked the underlying APIs. TsFileReader::queryByRow supports offset/limit, but the public tree query API cannot accept start/end at the same time. With a time range, the implementation still has to use query and apply the row window in the CLI. The table executor has a combined overload, but the reader wrapper does not expose it yet. I have kept the conservative condition for now; a follow-up API extension should add the pushdown together with out-of-range and exact-output tests.

cpp/tools/format/result_set_format.cc
ColinLeeo: This is the ResultSet serialization stage; the reader's time/TAG filters have already been applied before the ResultSet is created. emit_result_set calls is_null() to preserve the distinction between NULL and an empty string. Adding another reader filter in the formatter would change query semantics, so no reader filter API is needed here.

PR #919, opened 2026-08-25
fix(python): make dataframe index opt-in

ColinLeeo:
## Description

This PR makes the Python `TsFileDataFrame` persistent dataset index opt-in.

By default, `TsFileDataFrame(..., use_index=False)` now keeps the previous direct metadata loading behavior and does not create or reuse any `.tsidx` file. Users can explicitly enable persistent indexing with `use_index=True`; when enabled, the dataframe will reuse a matching index if present, or build one when the matching index is missing or stale.

The persistent index file name is now scoped by the expanded canonical TsFile set, so different subsets under the same directory no longer contend for the same `.tsidx` file.

## Changes

- Add `use_index: bool = False` to `TsFileDataFrame`.
- Preserve direct scan behavior when indexing is disabled.
- Use a hash of the expanded absolute file set in the persistent index file name.
- Keep index validation based on file fingerprints before reuse.
- Add tests for default no-index behavior.
- Add tests for same-directory different-file-subset index isolation.
- Update runtime/index-specific tests to pass `use_index=True`.

## Validation

- `./mvnw -P with-cpp -pl cpp -DskipTests package`
- `./mvnw -P with-python -pl python test`
  - `200 passed`

### DIFF
--- a/python/tests/test_dataset_index.py
+++ b/python/tests/test_dataset_index.py
@@ -244,14 +244,14 @@ def _write_runtime_devices_file(path):
 def test_hot_construction_maps_index_without_opening_readers(tmp_path, monkeypatch):
     source = tmp_path / "part.tsfile"
     _write_runtime_file(source, 0)
-    with TsFileDataFrame(str(source), show_progress=False) as first:
+    with TsFileDataFrame(str(source), show_progress=False, use_index=True) as first:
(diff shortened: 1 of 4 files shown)


hongzhi-gao (commented):
LGTM

PR #926, opened 2026-08-29
fix(cpp): report the stored codec in reader schema, not library defaults

kkzi:
## Problem

`TsFileReader::get_timeseries_schema()` reports an encoding and a compressor for
every series it returns, but those values never came from the file. Each
`MeasurementSchema` was built with the two-argument constructor, which fills in
`common::get_value_encoder(data_type)` and `common::get_default_compressor()`.

So a series written as `PLAIN`/`UNCOMPRESSED` was reported as `TS_2DIFF`/`LZ4`:
the schema described the library defaults, not the file. Any caller that
inspects a file to learn how it was encoded — a converter, a compatibility
check, a `tsfile`-inspecting tool — got an answer that happened to be right only
when the writer used the defaults.

## Why the metadata index cannot supply it

Encoding and compression are per-chunk properties, and on disk they live only in
the chunk header. `ChunkMeta::deserialize_from()` reads nothing but
`offset_of_chunk_header_` (plus an optional statistic), so `ChunkMeta`'s own
`encoding_`/`compression_type_` fields are never assigned — and the default
constructor does not initialize them. Reading them from a `ChunkMeta` obtained
via the metadata index reads uninitialized memory.

The fix therefore reads the chunk header back from the file at that offset and
deserializes it, which is the same approach `ChunkReader::load_by_meta()`
already uses. `TsFileIOReader::get_read_file()` is added for the raw byte
access. If the header cannot be read the previous defaults are used, so schema
introspection never fails outright.

The read buffer is sized from the measurement name length rather than being a
fixed size: the name sits between the chunk type byte and the codec bytes in the
chunk header, so a long enough name would push the codecs past a fixed buffer
and silently fall back to the defaults — the exact bug being fixed. The
600-character-name test below covers this.

## Tests

`cpp/test/reader/timeseries_schema_codec_test.cc`, four tests. Every series is
registered with codecs that differ from the library defaults for its data type
(INT32 defaults to `TS_2DIFF`, FLOAT to `GORILLA`, and the default compressor is
`LZ4` when compiled in), so a reader that reports defaults is caught:

- `ReportsStoredEncodingAndCompressor` — non-aligned device, `PLAIN`/`UNCOMPRESSED`
  INT32 and `TS_2DIFF`/`UNCOMPRESSED` FLOAT.
- `ReportsStoredCodecForAlignedDevice` — the aligned write path; also asserts the
  data type is the real one, not the `VECTOR` placeholder.
- `ReportsCodecPerSeriesNotPerDevice` — one device mixing `UNCOMPRESSED` and
  `SNAPPY`, so an implementation that hard-coded `UNCOMPRESSED` cannot pass.
  Gated on `ENABLE_SNAPPY`.
- `ReportsStoredCodecForLongMeasurementName` — a 600-character measurement name.

## Verification

- The test commit is checked out on its own, builds, and fails 4/4 — reporting
  `TS_2DIFF`/`LZ4` where the file stores `PLAIN`/`UNCOMPRESSED`. The fix commit
  passes 4/4.
- Full suite on MSVC Debug: 790 tests, 787 passed, 3 skipped (environment-gated),
  0 failed — exactly +4 over the baseline.
- Each test was confirmed non-vacuous: temporarily reverting to the two-argument
  constructor fails the first three; temporarily restoring a fixed 256-byte
  header buffer fails only the long-name test.
- clang-format 17.0.6 clean (matching the pinned `clang.format.version`).
- cppcheck with this repo's CI flag set reports no findings in either changed
  file.

Built and run on MSVC Debug locally; the rest of the `unit-test-cpp` matrix
(Linux/macOS, ASan, Release) is left to CI.

### DIFF
--- a/cpp/src/reader/tsfile_reader.cc
+++ b/cpp/src/reader/tsfile_reader.cc
@@ -40,42 +40,6 @@ struct DeviceMetaEntry {
     int64_t end_offset;
 };
 
-int read_chunk_header_codec(ReadFile* read_file, int64_t chunk_header_offset,
-                            size_t measurement_name_len,
-                            common::TSEncoding& encoding,
-                            common::CompressionType& compression) {
-    if (read_file == nullptr || chunk_header_offset < 0) {
-        return E_INVALID_ARG;
-    }
-
-    // A chunk header contains the measurement name before its codec fields.
-    // Read enough bytes for the known name instead of relying on a fixed-size
-    // header buffer.
-    constexpr size_t kMaxVarIntLen = 5;
-    std::vector<char> buffer(1 + kMaxVarIntLen + measurement_name_len +
-                             kMaxVarIntLen + 3);
-    int32_t read_len = 0;
-    int ret = read_file->read(chunk_header_offset, buffer.data(),
-                              static_cast<int32_t>(buffer.size()), read_len);
-    if (ret != E_OK) {
-        return ret;
-    }
-    if (read_len < ChunkHeader::MIN_SERIALIZED_SIZE) {
-        return E_TSFILE_CORRUPTED;
-    }
-
-    common::ByteStream input;
-    input.wrap_from(buffer.data(), read_len);
-    ChunkHeader chunk_header;
-    ret = chunk_header.deserialize_from(input);
-    if (ret != E_OK) {
-        return ret;
-    }
-    encoding = chunk_header.encoding_type_;
-    compression = chunk_header.compression_type_;
-    return E_OK;
-}
-
 int parse_paths(const std::vector<std::string>& path_list,
                 std::vector<Path>& parsed_paths) {
     try {
--- a/cpp/src/reader/tsfile_reader.cc
+++ b/cpp/src/reader/tsfile_reader.cc
@@ -40,6 +40,42 @@ struct DeviceMetaEntry {
     int64_t end_offset;
 };
 
+int read_chunk_header_codec(ReadFile* read_file, int64_t chunk_header_offset,
+                            size_t measurement_name_len,
+                            common::TSEncoding& encoding,
+                            common::CompressionType& compression) {
+    if (read_file == nullptr || chunk_header_offset < 0) {
+        return E_INVALID_ARG;
+    }
+
+    // A chunk header contains the measurement name before its codec fields.
+    // Read enough bytes for the known name instead of relying on a fixed-size
+    // header buffer.
+    constexpr size_t kMaxVarIntLen = 5;
+    std::vector<char> buffer(1 + kMaxVarIntLen + measurement_name_len +
+                             kMaxVarIntLen + 3);
+    int32_t read_len = 0;
+    int ret = read_file->read(chunk_header_offset, buffer.data(),
+                              static_cast<int32_t>(buffer.size()), read_len);
+    if (ret != E_OK) {
+        return ret;
+    }
+    if (read_len < ChunkHeader::MIN_SERIALIZED_SIZE) {
+        return E_TSFILE_CORRUPTED;
+    }
+
+    common::ByteStream input;
+    input.wrap_from(buffer.data(), read_len);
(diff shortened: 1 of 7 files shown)


cpp/src/reader/tsfile_reader.cc
```diff
@@ -471,8 +519,33 @@ int TsFileReader::get_timeseries_schema(
...
+                // read: schema introspection must not fail the whole call.
+                if (read_chunk_header_codec(
+                        tsfile_executor_->get_tsfile_io_reader()
+                            ->get_read_file(),
+                        chunk_meta_list->front()->offset_of_chunk_header_,
+                        measurement_name.size(), stored_enc,
+                        stored_comp) == E_OK) {
+                    enc = stored_enc;
+                    comp = stored_comp;
+                }
```
ColinLeeo: In java, the implement is 
```
 /** Get measurement schema by chunkMetadatas. */
  public MeasurementSchema getMeasurementSchema(List<IChunkMetadata> chunkMetadataList)
      throws IOException {
    if (chunkMetadataList.isEmpty()) {
      return null;
    }
    IChunkMetadata lastChunkMetadata = chunkMetadataList.get(chunkMetadataList.size() - 1);
    ChunkHeader header = readChunkHeader(lastChunkMetadata.getOffsetOfChunkHeader(), null);
    return new MeasurementSchema(
        lastChunkMetadata.getMeasurementUid(),
        header.getDataType(),
        header.getEncodingType(),
        header.getCompressionType());
  }
```

TsFile format allows different chunks of the same timeseries to use different encodings and compression types. Therefore, selecting a chunk here is an API behavior choice rather than a file-format invariant.

For consistency with the Java implementation, could we use the last ChunkMeta instead of front()? 

TsFileSequenceReader#getMeasurementSchema() reads the header referenced by the last IChunkMetadata:
IChunkMetadata lastChunkMetadata =
    chunkMetadataList.get(chunkMetadataList.size() - 1);
ChunkHeader header =
    readChunkHeader(lastChunkMetadata.getOffsetOfChunkHeader(), null);

Aligning the C++ behavior with Java would avoid returning different schemas for the same valid TsFile.

ColinLeeo:
LGTM overall.

cpp/src/reader/tsfile_reader.cc
kkzi: Thanks for pointing this out. The implementation has been updated in commit `aadeec5`.
 
`TsFileReader::get_timeseries_schema()` now reads the encoding and compression type from the last chunk header instead of the first chunk metadata entry, matching the behavior of the Java implementation. For aligned time series, the value chunk metadata list is used.
 
I also updated recovery writing to retain the codec information from the latest recovered chunk, fixed codec fields in `ChunkMeta` initialization and cloning, and added a regression test covering different codecs across multiple chunks.
 
The modified files were formatted with clang-format 17.0.6, and the relevant source compilation and existing schema tests passed.

cpp/src/reader/tsfile_reader.cc
```diff
@@ -35,6 +40,42 @@ struct DeviceMetaEntry {
...
+    ChunkHeader chunk_header;
+    ret = chunk_header.deserialize_from(input);
+    if (ret != E_OK) {
+        return ret;
+    }
+    encoding = chunk_header.encoding_type_;
+    compression = chunk_header.compression_type_;
+    return E_OK;
+}
+
```
ColinLeeo: Duplicate code ?

kkzi: You're right — commit `aadeec5` accidentally left a second, unreferenced copy of `read_chunk_header_codec()` at file scope, while the updated copy inside `namespace storage` is the one `get_timeseries_schema()` actually resolves to. Removed the dead copy in `d782b44`; no behavior change. Verified with clang-format 17.0.6 and the `TsFileReaderTest` suite (20/20 passed).

ColinLeeo (approved):
LGTM

PR #927, opened 2026-08-29
fix(cpp): open non-ASCII file paths through the wide CRT on Windows

kkzi:
## Problem

On Windows, every file open in the C++ implementation goes through the narrow CRT `open()`/`_open()`. That function interprets its `char*` argument in the process's **active code page**, not UTF-8. TsFile paths are UTF-8 everywhere else in the API, so any path containing a non-ASCII character is silently reinterpreted:

- `WriteFile` creates a file whose on-disk name is mojibake rather than the requested path.
- `ReadFile` cannot open a file that was written with the same UTF-8 path.
- Worst case, `RestorableTsFileIOWriter`'s self-check opens the target to decide whether a previous run crashed. When the open fails on a non-ASCII path, the check concludes the file is **incomplete and rewritable** — a complete TsFile is reported as `has_crashed() == true` / `can_write() == true`, so the writer truncates and overwrites it. That is silent data loss, not just an I/O error.

None of this surfaces as a crash or an error code at the API boundary, which is why it went unnoticed: the write "succeeds", just under a different name.

## Fix

Add `cpp/src/file/utf8_file_open.h` with a single helper, `file_internal::open_utf8()`:

- On Windows, convert UTF-8 → UTF-16 with `MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, ...)` and open through `_wopen()`, which takes the wide path and bypasses the code-page conversion entirely.
- Empty paths, paths with embedded NULs, and invalid UTF-8 are rejected up front with `errno` set, so callers keep their existing error contract instead of receiving a half-converted path.
- On POSIX the helper forwards directly to `open()`, so there is no behavior change and no new code path off Windows.

The three call sites — `read_file.cc`, `write_file.cc`, and `restorable_tsfile_io_writer.cc` — are routed through it. That is the whole change; nothing else about the I/O layer moves.

## Tests

`cpp/test/file/utf8_path_test.cc`, 5 tests in `Utf8PathTest`:

| Test | What it pins down |
|---|---|
| `WriteFileCreatesUtf8Path` | the created file's real on-disk name equals the requested UTF-8 path |
| `ReadFileOpensUtf8Path` | a non-ASCII path can be opened for reading |
| `Utf8PathRoundTripsBetweenWriteAndRead` | write-then-read with the same UTF-8 path resolves to the same file |
| `RestorableWriterSelfChecksUtf8Path` | a complete file on a non-ASCII path is **not** reported as crashed/rewritable — the data-loss case above |
| `OpenUtf8RejectsInvalidPaths` | empty, embedded-NUL, and invalid-UTF-8 inputs are rejected with `errno`, not silently mangled |

The tests use non-ASCII path components, so on a machine whose active code page is not UTF-8 they exercise exactly the mismatch described above.

## Verification

Commits are split so the defect is demonstrable rather than asserted:

- At the **test commit** alone (`8415e18`), 4/4 fail. `RestorableWriterSelfChecksUtf8Path` fails with `has_crashed()` actual `true` / expected `false` and `can_write()` actual `true` / expected `false` — the data-loss signature reproduced directly.
- At the **fix commit** (`c0484bc`), 5/5 pass. (The fifth test targets `open_utf8()`'s rejection behavior, so it only exists once the helper does.)
- Full C++ suite on MSVC 2019 x64 Debug at the PR base: **808 tests, 805 passed, 3 skipped, 0 failed** — the 3 skips are the pre-existing fixture/external-index tests.
- `cppcheck` with the exact flag set from `cppcheck.yml`: **no findings in any file this PR touches**, and a tree-wide run diffed against the PR base is byte-identical, so nothing new is introduced.
- `clang-format` 17.0.6 (the pinned `clang.format.version`) reports all 5 files clean; all files are LF-only and carry the Apache license header.

Only the MSVC Debug configuration was exercised locally. The remaining `unit-test-cpp` combinations — Linux and macOS, gcc and clang, ASan/UBSan, Release — are left to CI. On POSIX the helper is a direct passthrough to `open()`, so no behavior change is expected there.

### DIFF
--- a/cpp/src/cwrapper/tsfile_cwrapper.cc
+++ b/cpp/src/cwrapper/tsfile_cwrapper.cc
@@ -19,9 +19,9 @@
 
 #include "cwrapper/tsfile_cwrapper.h"
 
+#include <fcntl.h>
 #include <file/write_file.h>
 #include <reader/qds_without_timegenerator.h>
-#include <sys/stat.h>
 #include <writer/tsfile_table_writer.h>
 
 #ifdef _WIN32
@@ -115,24 +115,18 @@ WriteFile write_file_new(const char* pathname, ERRNO* err_code) {
     int ret;
     init_tsfile_config();
 
-    struct stat path_stat {};
-    if (stat(pathname, &path_stat) == 0) {
-#ifdef _WIN32
-        const bool is_dir = (path_stat.st_mode & _S_IFDIR) != 0;
-#else
-        const bool is_dir = S_ISDIR(path_stat.st_mode);
-#endif
-        *err_code = is_dir ? common::E_FILE_OPEN_ERR : common::E_ALREADY_EXIST;
-        return nullptr;
-    }
-
-    int flags = O_RDWR | O_CREAT | O_TRUNC;
+    int flags = O_RDWR | O_CREAT | O_TRUNC | O_EXCL;
 #ifdef _WIN32
     flags |= O_BINARY;
 #endif
     mode_t mode = 0666;
     storage::WriteFile* file = new storage::WriteFile;
     ret = file->create(pathname, flags, mode);
+    if (ret != common::E_OK) {
+        delete file;
+        *err_code = ret;
+        return nullptr;
+    }
     *err_code = ret;
     return file;
 }
--- a/cpp/src/file/read_file.cc
+++ b/cpp/src/file/read_file.cc
@@ -167,8 +167,10 @@ int ReadFile::open(const std::string& file_path) {
     // non-regular paths so both platforms report the same stable error code.
     // If stat itself fails, keep the normal open() path so missing or
     // inaccessible files continue to report E_FILE_OPEN_ERR.
+    std::wstring wide_path;
     struct __stat64 preopen_stat;
-    if (_stat64(file_path_.c_str(), &preopen_stat) == 0 &&
+    if (file_internal::utf8_to_wide(file_path_, wide_path) &&
+        _wstat64(wide_path.c_str(), &preopen_stat) == 0 &&
         (preopen_stat.st_mode & _S_IFMT) != _S_IFREG) {
         return E_INVALID_PATH;
     }
--- a/cpp/src/file/read_file.cc
+++ b/cpp/src/file/read_file.cc
@@ -38,6 +38,7 @@ ssize_t pread(int fd, void* buf, size_t count, uint64_t offset);
 #include "common/global.h"
 #include "common/logger/elog.h"
 #include "common/tsfile_common.h"
+#include "file/utf8_file_open.h"
 #include "utils/injection.h"
 #include "utils/util_define.h"  // ssize_t and other platform-compat shims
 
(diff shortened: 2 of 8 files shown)


kkzi:
The ubuntu/macos failures on the previous head were **not caused by the code change** — they were a test race, now fixed in `7eabec4`.

## Root cause

`gtest_discover_tests` registers every gtest case as its own ctest entry, so `ctest --parallel 3` runs several cases of the same binary as **separate processes in one shared working directory**. The four file-based tests shared a single fixed filename and unlinked it unconditionally in `SetUp`/`TearDown`, so a concurrently starting case deleted the file another case had just created.

The ubuntu log shows the interleave directly:

```
Start 416: Utf8PathTest.Utf8PathRoundTripsBetweenWriteAndRead
413/829 Test #414: Utf8PathTest.WriteFileCreatesUtf8Path ... Passed
Start 417: Utf8PathTest.RestorableWriterSelfChecksUtf8Path   <- SetUp unlinks
414/829 Test #415: Utf8PathTest.ReadFileOpensUtf8Path ***Failed  <- stat: ENOENT
```

The scheduling order decides which case loses, which is why the two platforms failed differently (ubuntu: `stat` ENOENT right after creating the fixture; macos: `ReadFile::open` raced a concurrent truncate/unlink so the size/magic check returned `E_TSFILE_CORRUPTED`, plus the round-trip test's `stat` ENOENT) and why the Windows legs passed by luck.

## Fix

Each case now derives its own filename from the gtest case name (appended to the non-ASCII stem in both the UTF-8 and UTF-16 spellings, so the wide-path assertion still checks the intended name). Every process only ever creates/removes its own file.

Verified locally by reproducing the CI pattern — the four cases run as four concurrent processes in one directory, five rounds, all green; full suite green; clang-format clean.

ColinLeeo:
Thanks for your pr.

Could we update the existing-file checks together with this change?

TsFileWriter::open() still checks the path with narrow access(), and write_file_new() does the same with narrow stat(). On Windows with a non-UTF-8 active code page, those checks may look up a mojibake path and report that an existing non-ASCII file does not exist. The subsequent open_utf8() then resolves the intended Unicode path and opens it with O_TRUNC, silently truncating the existing file instead of returning E_ALREADY_EXIST.

Please make the existence checks use the same UTF-8-to-UTF-16 path conversion, or preferably replace the check-then-open sequence with an atomic O_EXCL open. It would also be good to add regression tests through TsFileWriter::open() and write_file_new() verifying that an existing Unicode-path file is rejected and its contents remain unchanged.

kkzi:
Thanks for pointing this out. I updated the PR accordingly.

- Replaced the check-then-open existence checks in `TsFileWriter::open()` and `write_file_new()` with atomic `O_CREAT | O_EXCL` creation.
- Mapped `EEXIST` to `E_ALREADY_EXIST` in `WriteFile::do_create()`.
- Added regression tests for both APIs using an existing Unicode path; both verify that the original file contents remain unchanged.
- Rebased onto the latest `develop` and resolved the `read_file.cc` conflict while preserving the new read backend. The Windows directory preflight now uses the same UTF-8-to-UTF-16 conversion.

Validation:
- MSVC Debug build: passed.
- Focused tests: 8 passed (`Utf8PathTest.*` and `CReleaseTest.TestCreateFile`).
- Full MSVC Debug suite: 926 passed, 3 skipped.
- clang-format 17.0.6: passed.
- cppcheck 2.21.0 baseline comparison: no new diagnostics.

ColinLeeo (approved):
LGTM

PR #928, opened 2026-08-31
Add initial Go API bindings

ColinLeeo:
## Summary

- add a cgo-based Go module with Reader, ResultSet, Writer, Tablet, tree/table schemas, typed values, and error mapping
- expose stable public C writer and exact tree-query entry points without depending on private Python symbols
- add tree/table round-trip tests, lifetime and race coverage, examples, build documentation, and Linux/macOS CI

## Validation

- cmake --build cpp/target/build -j4
- 25 C wrapper tests passed
- go vet ./...
- go test ./...
- go test -race ./...
- GOEXPERIMENT=cgocheck2 go test ./...
- clang-format --dry-run --Werror on changed C/C++ files
- git diff --check

## Scope

This initial API targets Linux amd64 and macOS arm64 with cgo enabled. The native TsFile C++ library remains an explicit prerequisite; the Go module does not install native dependencies. Windows, a pure-Go implementation, Arrow, dataframe APIs, and tag-filter construction are intentionally out of scope.

### DIFF
--- a/.github/workflows/unit-test-go.yml
+++ b/.github/workflows/unit-test-go.yml
@@ -1,0 +1,60 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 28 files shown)


ColinLeeo:
Superseded by #929, which is based directly on the latest develop branch and excludes unrelated Python dependency commits.

PR #929, opened 2026-08-31
Add initial Go API bindings

ColinLeeo:
## Summary

- add a cgo-based Go module with Reader, ResultSet, Writer, Tablet, tree/table schemas, typed values, and error mapping
- expose stable public C writer and exact tree-query entry points without depending on private Python symbols
- add tree/table round-trip tests, lifetime and race coverage, examples, build documentation, and Linux/macOS CI

## Validation

- cmake --build cpp/target/build -j4
- 25 C wrapper tests passed
- go vet ./...
- go test -count=1 ./...
- go test -race -count=1 ./...
- GOEXPERIMENT=cgocheck2 go test -count=1 ./...
- clang-format --dry-run --Werror on changed C/C++ files
- git diff --check

## Scope

This initial API targets Linux amd64 and macOS arm64 with cgo enabled. The native TsFile C++ library remains an explicit prerequisite; the Go module does not install native dependencies. Windows, a pure-Go implementation, Arrow, dataframe APIs, and tag-filter construction are intentionally out of scope.


docs/superpowers/specs/2026-08-27-go-api-design.md
```diff
@@ -0,0 +1,313 @@
+# TsFile Go API Design
```
HTHou: docs/superpowers?

CritasWang (commented):
## Code Review — PR #929: Add initial Go API bindings / 代码评审

### Verdict / 结论

**LGTM — 整体质量很高，没有阻塞性问题，可以合入。** / Overall very high quality, no blocking issues, good to merge. All findings below are non-blocking suggestions.

### What I did / 我做了什么

- Checked out the PR head locally (`feature/go-api-support` @ `2ef74d1b`) and reviewed all 24 changed files (Go module, C wrapper, tests, CI, docs).
- GitHub CI is fully green: cppcheck, ASan, C++ unit-test matrix (Linux/macOS/Windows), CodeQL, and the new Go workflow (full `go vet` + `go test` + `go test -race` on Linux/macOS and Windows MinGW).
- Local verification against a locally built `libtsfile.dylib`: `go vet ./...`, `go test -count=1 ./...`, `go test -race`, `GOEXPERIMENT=cgocheck2 go test` all pass; both examples write and read back real TsFiles. (My local `./mvnw -P with-cpp package` failed at the gtest compile step because this machine's SDK lacks `cxxabi.h` — an environment issue unrelated to this PR.)

### Highlights / 亮点

- cgo hygiene is excellent: every Go string/byte slice is copied into C memory before the call and freed right after; Go pointers are never retained by C; all cgo traffic is confined to one auditable file.
- Lock ordering is consistent (reader → result set); `Close` is idempotent with a well-documented retry-on-failure ownership model.
- Enum and error-code pinning tests against `errno_define_c.h` / `tsfile_cwrapper.h` prevent silent ABI drift.
- The C wrapper changes are purely additive, null-guarded delegates; no existing C/C++/Python behavior changes.
- The header doc fix (0-based → 1-based column index) is correct: native `ResultSet::get_value(uint32_t)` is 1-based and the Go bridge's `column+1` matches.

### Issues & suggestions / 问题与建议（non-blocking）

1. **`AddProperty` empty value becomes a NULL property** — semantic gap, see inline comment.
2. **Writer construction mutates a process-global config** — `WithMemoryThreshold` is effectively global and concurrent `NewWriter` races, see inline comment.
3. **`DataTypeBlob` is a dead end** — accepted by schema validation but no `SetBlob` setter, see inline comment.
4. **Design spec acceptance criteria not fully met + doc drift** — no cross-language tests, not all data types/multiple devices/empty ranges covered; spec still says "no Windows" while CI now covers it, see inline comment.
5. **MSVC CI only verifies artifacts** — Go tests run on MinGW only; PR summary wording can be more precise, see inline comment.
6. **(Pre-existing) `tsfile_result_set_metadata_get_column_name(meta, 0)` would read `column_names[-1]`** — consider adding the lower bound now that the docs pin `[1, column_num]`, see inline comment.
7. **Silent post-close behavior** — `ResultSet.Metadata()` returns nil and `Tablet.Rows()` returns 0 after close without error/documentation, see inline comment.
8. **`Makefile` fmt target hardcodes the file list** — see inline comment.
9. (Minor) The new `tsfile_reader_query_tree` catches only `std::bad_alloc`, consistent with existing style; a future hardening pass could catch all C++ exceptions uniformly across the C ABI.

Details are in the inline comments below. / 详见下方行内评论。

go/tsfile/writer.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// WriteTreeTablet writes a tree-model tablet without consuming it.
+func (w *Writer) WriteTreeTablet(tablet *Tablet) error { return w.writeTablet(tablet, false) }
+
+// WriteTableTablet writes a table-model tablet without consuming it.
+func (w *Writer) WriteTableTablet(tablet *Tablet) error { return w.writeTablet(tablet, true) }
+
+// AddProperty adds a binary TsFile property.
+func (w *Writer) AddProperty(key string, value []byte) error {
```
CritasWang, line 157: **建议 / Suggestion (non-blocking)**

中文：`AddProperty(key, []byte{})` 时 `copyCBytes` 返回 `(nil, 0)`，native 端 `TsFilePropertyValue(data == nullptr)` 会把它存成 **NULL property** 而不是空值 —— Go API 目前无法表达“空值 vs NULL”的区别。建议在文档中说明该语义，或直接拒绝空 value。

English: With an empty value, `copyCBytes` returns `(nil, 0)` and the native `TsFilePropertyValue(data == nullptr)` stores a **NULL property**, not an empty one — the Go API cannot express "empty vs NULL". Please document this, or reject empty values.

cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -1959,6 +1989,165 @@ ERRNO _tsfile_writer_add_tsfile_property(TsFileWriter writer, const char* key,
     }
 }
 
+TsFileGenericWriter tsfile_generic_writer_new(const char* pathname,
```
CritasWang, line 1992: **建议 / Suggestion (non-blocking)**

中文：这个公开入口委托给 `_tsfile_writer_new`，后者会把 `memory_threshold` 写入进程全局 `common::g_config_value_.chunk_group_size_threshold_`（继承自 Python 路径的全局副作用）。因此 Go 的 `WithMemoryThreshold` 实际上是进程级生效而非 per-writer，且并发 `NewWriter` 会对该全局产生数据竞争。建议在 Go 层把 writer 构造串行化（如 `sync.Mutex` 包住构造函数），或在 README 中明确这一限制。

English: This entry delegates to `_tsfile_writer_new`, which writes `memory_threshold` into the process-global `common::g_config_value_.chunk_group_size_threshold_` (a global side effect inherited from the Python path). So Go's `WithMemoryThreshold` is effectively process-wide rather than per-writer, and concurrent `NewWriter` calls race on that global. Consider serializing writer construction in Go (e.g. a mutex around the constructor), or documenting this limitation in the README.

go/tsfile/schema.go
```diff
@@ -0,0 +1,110 @@
...
+
+func columnNames(columns []ColumnSchema) []string {
+	names := make([]string, len(columns))
+	for i := range columns {
+		names[i] = columns[i].Name
+	}
+	return names
+}
+
+func validDataType(value DataType) bool {
```
CritasWang, line 78: **建议 / Suggestion (non-blocking)**

中文：`validDataType` 接受 `DataTypeBlob`，可以把它注册进 table/timeseries schema，但 `Tablet` 没有对应的 `SetBlob`，blob 列永远写不了，是个死胡同。建议暂时把 Blob 从合法集合中移除，或补一个 setter。

English: `validDataType` accepts `DataTypeBlob`, so it can be registered in a schema, but `Tablet` has no `SetBlob` — blob columns can never be written, a dead end. Either drop Blob from the valid set for now, or add the missing setter.

cpp/src/cwrapper/tsfile_cwrapper.h
```diff
@@ -903,7 +919,7 @@ ResultSetMetaData tsfile_result_set_get_metadata(ResultSet result_set);
 /**
  * @brief Gets column name by index from metadata.
  *
- * @param column_index [in] Column position (0 ≤ index < column_num).
+ * @param column_index [in] Column position in [1, column_num].
```
CritasWang, line 922: **建议 / Suggestion (pre-existing, non-blocking)**

中文：预存在问题（非本 PR 引入）：`tsfile_result_set_metadata_get_column_name` 只做上界检查（`column_index > column_num`），传 `0` 会读 `column_names[-1]`。既然文档已统一为 `[1, column_num]`，建议顺手补下界校验（`column_index < 1` 返回 NULL），这样 Go 桥即使出错也不会踩到越界。

English: Pre-existing (not introduced here): `tsfile_result_set_metadata_get_column_name` only checks the upper bound (`column_index > column_num`); passing `0` reads `column_names[-1]`. Now that the docs pin `[1, column_num]`, consider adding the lower-bound check (`column_index < 1` → NULL) as well, so the Go bridge can't hit it even on a bug.

.github/workflows/unit-test-go.yml
```diff
@@ -0,0 +1,117 @@
...
+          "C:\mingw64\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          Write-Host "Added native library paths: $nativeLib and C:\mingw64\bin"
+      - name: Test Go API
+        if: matrix.toolchain != 'msvc'
+        working-directory: go
+        run: |
+          go vet ./...
+          go test ./...
+          go test -race ./...
+      - name: Verify MSVC native build
```
CritasWang, line 108: **建议 / Suggestion (non-blocking)**

中文：MSVC job 只校验 `tsfile.dll` / `tsfile.lib` 产物，Go 测试实际只跑在 MinGW 下。这个安排本身合理（MSVC 下的 cgo 支持属于后续工作），但 PR 摘要里 “Linux/macOS CI” 的说法与事实不一致 —— Windows MinGW 其实跑了全套 `go vet`/test/race。建议把摘要写准确：Linux/macOS + Windows MinGW 跑全套 Go 测试，MSVC 仅验证原生产物。

English: The MSVC job only verifies that `tsfile.dll` / `tsfile.lib` are produced; the Go tests actually run under MinGW only. That split is reasonable (cgo on MSVC can be follow-up work), but the PR summary's "Linux/macOS CI" understates reality — Windows MinGW runs the full `go vet` / `go test` / `go test -race` suite. Consider making the summary precise: full Go tests on Linux/macOS and Windows MinGW, MSVC verifies native artifacts only.

docs/superpowers/specs/2026-08-27-go-api-design.md
```diff
@@ -0,0 +1,313 @@
...
+- flush followed by close; and
+- reader/result-set parent-child lifetime behavior.
+
+Cross-language tests read Python or Java fixtures from Go and read Go-written
+files with an existing Python or C++ reader. C wrapper lifetime tests run under
+ASan on supported CI jobs, with valgrind as an optional Linux supplement.
+`go test -race` exercises concurrent method and close calls to verify internal
+serialization.
+
+Acceptance requires:
```
CritasWang: **建议 / Suggestion (non-blocking)**

中文：设计文档的验收标准还没有完全兑现：本 PR 的集成测试只覆盖 Int64/String/Double/Boolean、单设备，缺跨语言 round-trip、全部数据类型、多设备、空时间范围查询。另外第 22 行仍写 “does not include … Windows”，但实现已经加了 Windows MinGW/MSVC CI。建议更新 spec：标注哪些项已满足、哪些留作 follow-up issue，避免文档与实现漂移。

English: The spec's acceptance criteria aren't fully met yet: the integration tests cover only Int64/String/Double/Boolean on a single device — no cross-language round-trips, not all data types, no multiple devices, no empty time-range queries. Also, line 22 still says "does not include … Windows" while this PR adds Windows MinGW/MSVC CI. Consider updating the spec: mark what's satisfied and file follow-ups for the rest, to keep docs and implementation in sync.

go/tsfile/result_set.go
```diff
@@ -0,0 +1,181 @@
...
+		var err error
+		next, err = rs.handle.next()
+		rs.current = next
+		return err
+	})
+	return next, err
+}
+
+// Metadata returns a copy of the result columns, including timestamp at zero.
+func (rs *ResultSet) Metadata() []ColumnMetadata {
```
CritasWang, line 68: **建议 / Suggestion (non-blocking)**

中文：`ResultSet.Metadata()` 在关闭后静默返回 nil，`Tablet.Rows()`（tablet.go:150）在关闭后返回 0，都没有错误返回、godoc 也没说明。建议至少在注释里写明“关闭后返回 nil/0”，或考虑改为返回 `([]ColumnMetadata, error)`。

English: `ResultSet.Metadata()` silently returns nil after close, and `Tablet.Rows()` (tablet.go:150) returns 0 after close — no error, and the godoc doesn't say so. At minimum document "returns nil/0 after close", or consider returning an error.

go/Makefile
```diff
@@ -0,0 +1,32 @@
...
+test:
+	GOCACHE=$(GOCACHE) go test ./...
+
+race:
+	GOCACHE=$(GOCACHE) go test -race ./...
+
+vet:
+	GOCACHE=$(GOCACHE) go vet ./...
+
+fmt:
```
CritasWang, line 31: **建议 / Suggestion (non-blocking, nit)**

中文：`fmt` 目标硬编码了文件列表，以后新增 Go 文件容易漏掉。建议改成 `gofmt -l . | xargs gofmt -w` 之类自动覆盖全部文件。

English: The `fmt` target hardcodes the file list, which will silently miss new files. Prefer something like `gofmt -l . | xargs gofmt -w` to cover everything.

go/tsfile/writer.go
ColinLeeo: Fixed in 3b08479c2. The Go API now rejects empty property values with ErrInvalidArgument and documents that the native ABI uses a nil pointer to represent NULL rather than an empty byte string.

cpp/src/cwrapper/tsfile_cwrapper.cc
ColinLeeo: Addressed in 3b08479c2 by serializing NewWriter native construction with a package-level mutex. This protects the current process-global memory-threshold configuration from concurrent constructor races. The limitation is also documented in the implementation comments.

go/tsfile/schema.go
ColinLeeo: Fixed in 3b08479c2. DataTypeBlob is no longer accepted by Go schema validation until a corresponding typed setter and round-trip coverage are available.

cpp/src/cwrapper/tsfile_cwrapper.h
ColinLeeo: Fixed in 3b08479c2. The existing C wrapper now rejects column_index < 1 as well as values above the column count, matching the documented 1-based ABI.

.github/workflows/unit-test-go.yml
ColinLeeo: Thanks for pointing this out. The CI behavior is intentional: Linux/macOS and Windows MinGW run the complete Go vet/test/race suite, while the MSVC job verifies the native DLL and import library artifacts. The PR description will use this precise wording.

docs/superpowers/specs/2026-08-27-go-api-design.md
ColinLeeo: Updated in 3b08479c2. The design spec now reflects Windows MinGW/MSVC coverage, documents the representative data types and scenarios currently tested, and marks cross-language fixtures, broader data-type coverage, multi-device cases, empty ranges, and MSVC cgo execution as follow-up work.

go/tsfile/result_set.go
ColinLeeo: Addressed in 3b08479c2 by documenting the existing post-close behavior: ResultSet.Metadata returns nil and Tablet.Rows returns zero after close.

go/Makefile
ColinLeeo: Fixed in 3b08479c2. The Makefile fmt target now discovers and formats every Go file under the module instead of maintaining a hard-coded list.

ColinLeeo:
Thanks for the thorough review and the LGTM.

I addressed the actionable suggestions in commit 3b08479c2:

- Empty Go property values are rejected with ErrInvalidArgument and documented.
- NewWriter construction is serialized to protect the current process-global native memory-threshold configuration.
- DataTypeBlob is excluded from Go schema validation until a typed setter is available.
- The C metadata accessor now validates the documented 1-based lower bound.
- Post-close behavior for Metadata and Rows is documented.
- The Go Makefile formats all Go files automatically.
- The design spec now reflects Windows MinGW/MSVC CI coverage and clearly marks broader cross-language/data-type/multi-device coverage as follow-up work.

The CI description is also clarified: Linux/macOS and Windows MinGW run the full Go vet/test/race suite; the MSVC job verifies the native DLL and import library artifacts.

The broader C ABI exception-handling cleanup remains a separate follow-up because the current public error enum has no generic internal-exception code, and introducing a new error code would be a wider ABI decision.

Local validation passes with `go vet ./...`, `go test -count=1 ./...`, `go test -race -count=1 ./...`, `clang-format --dry-run --Werror`, and `git diff --check`.

PR #934, opened 2026-09-02
perf(cpp): cache per-device schema check for repeated tablet writes (#885)

### LINKED ISSUE #885: Optimize TsFileWriter::do_check_schema CPU overhead for repeated wide-tablet writes
## Problem

  `TsFileWriter::do_check_schema` shows noticeable CPU overhead when writing repeated tablets with the same device and schema.

  In our workload, each writer repeatedly writes tablets for a fixed device with a fixed set of measurements. However,
  `write_tablet()` calls `do_check_schema()` for every tablet write. This appears to repeatedly perform schema lookup and
  measurement-name matching even though the schema has already been registered and does not change during the file lifecycle.

  ## Evidence

  Using Windows Performance Analyzer on a 60-second TsFile archive workload, the active processing window showed:

  - `TsFileArchive.dll!storage::TsFileWriter::do_check_schema<MeasurementNamesFromTablet>` as the top TsFileArchive function
  hotspot
  - `std::string::compare` also appeared as a related hotspot
  - The workload writes tablets with the same device and the same schema repeatedly
  - TsFile writing took longer than the input data duration

  Example hotspot:

  ```text
  TsFileArchive.dll!storage::TsFileWriter::do_check_schema<storage::MeasurementNamesFromTablet>
```
  ## Suspected Cause

  write_tablet() creates or looks up schema information on every call:

  - creates a StringArrayDeviceID from tablet.insert_target_name_
  - looks up the device in schemas_
  - iterates all measurement names
  - looks up each measurement in measurement_schema_map_
  - fills chunk_writers and data_types

  For wide tablets and high-frequency writes, this repeated per-tablet schema validation becomes expensive.

  ## Suggested Optimization

  Add a cached or prepared schema path for repeated tablet writes.

  Possible approaches:

  1. Cache the resolved schema for the last used device/tablet schema inside TsFileWriter.
  2. Add an explicit prepared API, for example:

  PreparedTabletSchema prepare_tablet_schema(device_id, schema_vec);
  int write_tablet_prepared(const Tablet& tablet, const PreparedTabletSchema& prepared);

  The cached/prepared data could include:

  - MeasurementSchemaGroup*
  - resolved ChunkWriter* list
  - resolved TSDataType list

  This would allow repeated writes with the same device and schema to skip per-column name lookup.

  ## Expected Benefit

  Reduce CPU overhead in high-throughput TsFile writing workloads, especially for wide schemas where the same tablet schema is
  written repeatedly.

  ## Notes

  This should preserve the existing validation behavior for dynamic schemas or changing devices. The optimization can be limited
  to cache hits where both device and schema identity match.

kkzi:
Title:
perf(cpp): cache per-device schema check for repeated tablet writes (#885)

Body:
## Problem

Closes #885.

`TsFileWriter::do_check_schema` / `do_check_schema_aligned` re-resolve every
measurement name against `measurement_schema_map_` on **every**
`write_tablet` / `write_tablet_aligned` / `write_record` call, even though
the device schema is registered once and does not change during the file
lifecycle. For wide tablets written repeatedly this per-column string lookup
is a top CPU hotspot (profiling in #885).

## Change

Cache the resolved chunk writers and data types per device in
`MeasurementSchemaGroup` (`cpp/src/common/schema.h`), resolved once and
reused while the tablet's measurement **name sequence** is unchanged:

- **Cache key is the name sequence, not the column count.** Cached entries
  are reused by position, so the same count with a different name order (or
  one column swapped) would silently write values into the wrong column with
  the wrong data type. A mismatch drops the stale cache and re-resolves, so
  dynamic schemas keep the existing validation behavior.
- **Plain and aligned paths keep separate caches** (a shared flag would let
  whichever path runs first lock the other out permanently).
- **Only fully-resolved results are cached.** A NULL chunk writer for a
  not-yet-registered measurement must not be pinned, or the column would
  stay masked even after it is registered later.
- Cached entries are the same **non-owning** pointers the uncached path
  returns: `chunk_writer_` is only freed in `destroy()`, flush only resets
  it, and `register_timeseries` returns `E_ALREADY_EXIST` for an existing
  name instead of replacing the object, so a cached pointer cannot dangle.
- The written file is byte-identical to the uncached path.

`do_check_schema_table` (table model) is unchanged; this targets the tree
model paths named in #885.

## Tests

New suite `SchemaCheckCacheTest` (`cpp/test/writer/tsfile_writer_schema_cache_test.cc`):

1. `RepeatedSameSchemaRoundTrip` — hit path, incl. after a flush seals and
   resets the chunk writers; asserts every row and column (no dropped rows).
2. `SameCountDifferentNamesAndOrder` — same column count with a swapped /
   reordered name set re-resolves; values land in the column their name says.
3. `ColumnRegisteredAfterFirstWriteIsNotMasked` — the fully-resolved-only
   guard: a column unregistered at first write is written once registered.
4. `AlignedRepeatedAndReorder` — aligned path: own cache, hit + reorder.
5. `MultiDeviceCachesIndependent` — two devices with identical measurement
   names, interleaved writes, no cross-wiring.

Full suite: 830 tests, 827 passed, 3 skipped (environment-only skips:
external-index and cross-language fixture generators), 0 failures.

## Benchmark

A/B microbenchmark (not committed — happy to add it to the tree if wanted):
one device, 200 INT32/PLAIN columns, 2000 tablet writes × 1 row each (the
few-rows-wide-tablet shape from the issue), parallel write disabled, best of
5 rounds, MSVC 2019 x64:

| build | uncached (develop) | cached (this PR) | speedup | saved ns/col |
|-------|--------------------|------------------|---------|--------------|
| Release | 33.4 µs/write | 19.0 µs/write | **1.76×** | ~72 |
| Debug | 246.0 µs/write | 150.0 µs/write | 1.64× | ~480 |

The schema check was ~43% of the total `write_tablet` time in this shape.
With more rows per tablet the end-to-end percentage shrinks (the saving is
per write, the encode cost is per point), but the per-column lookup remains
the dominant fixed cost per call — which is what #885 profiles.

## Notes

- The benchmark above is a scratch local run; a permanent benchmark harness
can follow the direction from the maintainer feedback in #885.

### DIFF
--- a/cpp/src/common/schema.h
+++ b/cpp/src/common/schema.h
@@ -28,6 +28,7 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+#include <vector>
 
 #include "common/db_common.h"
 #include "writer/time_chunk_writer.h"
@@ -175,6 +176,30 @@ struct MeasurementSchemaGroup {
     bool is_aligned_ = false;
     TimeChunkWriter* time_chunk_writer_ = nullptr;
     int64_t last_time_ = INT64_MIN;
+
+    // Per-device schema-check cache (do_check_schema /
+    // do_check_schema_aligned): the resolved chunk writers + data types for
+    // the last fully-resolved measurement-name sequence, so repeated writes
+    // with an unchanged schema skip the per-column measurement_schema_map_
+    // string lookup (#885). Entries are the SAME non-owning pointers the
+    // uncached path returns, so the written file is byte-identical.
+    //
+    // Guarded by the measurement NAME SEQUENCE, not just the column count:
+    // entries are reused by position, so the same count with a different
+    // name order (or one column swapped) would silently write values into
+    // the wrong column AND with the wrong data type.
+    //
+    // The plain and aligned paths keep separate caches: sharing one flag
+    // would let whichever path ran first lock the other out permanently.
+    std::vector<ChunkWriter*> cached_chunk_writers_;
+    std::vector<common::TSDataType> cached_data_types_;
+    std::vector<std::string> cached_measurement_names_;
+    bool schema_check_cached_ = false;
+
+    std::vector<ValueChunkWriter*> cached_value_chunk_writers_;
+    std::vector<common::TSDataType> cached_aligned_data_types_;
+    std::vector<std::string> cached_aligned_measurement_names_;
+    bool schema_check_aligned_cached_ = false;
 
     ~MeasurementSchemaGroup() {
         if (time_chunk_writer_ != nullptr) {
--- a/cpp/src/writer/tsfile_writer.cc
+++ b/cpp/src/writer/tsfile_writer.cc
@@ -468,10 +468,48 @@ int TsFileWriter::do_check_schema(
     }
     MeasurementSchemaMap& msm = device_schema->measurement_schema_map_;
     uint32_t measurement_count = measurement_names.get_count();
-    // chunk_writers.reserve(measurement_count);
+    // The getter is single-pass (next() advances an index), so buffer the
+    // name sequence up front. next() returns a reference, so this buffers
+    // pointers rather than copies, which lets the cache check below compare
+    // the real names and still fall through to the full lookup on mismatch.
+    static thread_local std::vector<const std::string*> names;
+    names.clear();
+    names.reserve(measurement_count);
     for (uint32_t i = 0; i < measurement_count; i++) {
-        auto ms_iter = msm.find(measurement_names.next());
+        names.push_back(&measurement_names.next());
+    }
+    // Column count alone is not a safe cache key: entries are reused by
+    // position, so the same count with a different name order (or one column
+    // swapped) would silently write values into the wrong column with the
+    // wrong data type.
+    if (device_schema->schema_check_cached_ &&
+        device_schema->cached_measurement_names_.size() == measurement_count) {
+        bool same_schema = true;
+        for (uint32_t i = 0; i < measurement_count; i++) {
+            if (device_schema->cached_measurement_names_[i] != *names[i]) {
+                same_schema = false;
+                break;
+            }
+        }
+        if (same_schema) {
+            for (uint32_t i = 0; i < measurement_count; i++) {
+                chunk_writers.push_back(
+                    device_schema->cached_chunk_writers_[i]);
+                data_types.push_back(device_schema->cached_data_types_[i]);
+            }
+            return E_OK;
+        }
+        // Schema changed for this device: drop the stale cache and re-resolve.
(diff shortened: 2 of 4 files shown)


ColinLeeo:
Thanks for the PR. This is a worthwhile optimization: avoiding repeated per-measurement schema lookups for recurring tablet layouts should improve write performance.

However, could we move this cache out of `MeasurementSchemaGroup` and make it a bounded, writer-level cache?

The current implementation adds six vectors and two flags to every device schema group. In my local 64-bit build, `sizeof(MeasurementSchemaGroup)` increases from 48 bytes to 208 bytes, which is about 160 bytes of fixed overhead per device before any cached names, writers, or data types are allocated. Table-model devices also pay this cost even though `do_check_schema_table()` does not use these cache fields.

I suggest maintaining a small bounded cache in `TsFileWriter`, with separate entries for the aligned and non-aligned tree-model paths. This keeps the additional memory bounded and avoids adding unused cache state to every registered device. For the single-device workload described in #885, a one-entry MRU/last-hit cache may already be sufficient; if benchmarks cover interleaved writes across multiple devices, a small LRU would avoid thrashing.

There is already a generic implementation in `cpp/src/common/cache/lru_cache.h`, but it does not appear ready for direct reuse. Some public APIs do not compile when instantiated, the reference-style lookup still copies the value, and the `MetadataQuerier` integration is currently inconsistent and inactive. We could either complete that implementation with tests and a non-copying lookup API, or introduce a focused writer-private bounded cache for this optimization.

kkzi:
Updated the schema-check cache implementation based on the review feedback.

Changes in commit `61bc506`:
- Moved schema-check cache state from `MeasurementSchemaGroup` to `TsFileWriter`.
- Added independently bounded one-entry MRU caches for non-aligned and aligned tree-model paths.
- Required exact device ID and measurement-name sequence matching for cache hits.
- Cached only fully resolved schemas; cached writer pointers remain non-owning.
- Added cache invalidation for writer initialization/destruction and successful timeseries registration.
- Kept table-model schema checking unchanged.
- Added regression coverage for flush reuse, reordered measurements, late registration, interleaved devices, aligned/non-aligned separation, and record paths.

Verification:
- `SchemaCheckCacheTest.*`: 6/6 passed
- `TsFileWriterTest.*`: 25/25 passed
- Full `TsFile_Test.exe`: 929 tests, 926 passed, 3 environment-dependent tests skipped, 0 failed
- `clang-format 17.0.6 --dry-run --Werror`: passed for all three changed files
- `cppcheck`: no new findings in the modified writer implementation; the repository-wide check remains affected by pre-existing baseline findings in unrelated files including `mem_alloc.cc` and `read_file.cc`.

PR #936, opened 2026-09-03
docs: add security policy

HTHou:
## Summary

- add ASF-compliant instructions for privately reporting undisclosed vulnerabilities
- define the TsFile security model and trust boundaries across the Java and C++ implementations, C API, Python and Go bindings, and command-line tools
- document expectations for untrusted files, resource exhaustion, optional runtime implementations, confidentiality, and integrity
- distinguish security-relevant findings from ordinary robustness issues

The policy follows the ASF vulnerability handling process and draws on the security guidance published by Apache Parquet, Arrow, ORC, CarbonData, and Avro.

## Testing

- `git diff --check`
- documentation-only change; no code tests required

PR #937, opened 2026-09-04
docs: clarify security issue classification

HTHou:
## Summary

- classify untrusted-file findings according to exploitability and demonstrated security-boundary impact
- clarify that crashes, assertion failures, out-of-memory conditions, and sanitizer findings without such impact are normally robustness issues
- retain case-by-case security treatment for disproportionate resource amplification across an actual untrusted boundary
- narrow the examples of security-relevant memory findings to exploitable corruption or out-of-bounds access that exposes unrelated process data

This follow-up refines the security model added in #936 and more clearly reflects the distinction used by comparable Apache data-format projects.

## Testing

- `git diff --check`
- documentation-only change; no code tests required

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -69,10 +69,10 @@ pointers or violating an API's ownership and lifetime requirements is outside th
 
 ### Reading Untrusted Files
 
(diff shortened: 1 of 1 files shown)


ColinLeeo (approved):
LGTM

PR #938, opened 2026-09-04
[Feature][C++][Python] Support seekable random-access TsFile sources

### LINKED ISSUE #930: [Feature][C++][Python] Support remote random-access reading for TsFile
## Motivation

Apache TsFile has been integrated into Hugging Face datasets, supporting
TsFile loading through:

```python
load_dataset("tsfile", data_files=...)
```

The required Python dependency has also been deployed in the Dataset Viewer
environment.

In Hugging Face streaming scenarios, datasets uses `xopen()` to obtain remote
file objects managed by fsspec. For Hugging Face and storage services that
support HTTP Range requests, these file objects provide random access to
specific byte ranges.

The Parquet Reader uses this capability to read the footer, metadata, and
required data blocks from remote files. TsFile also uses footer metadata and
offset-based reads, so it can follow the same remote random-access model.

Related Hugging Face work:

- https://github.com/huggingface/datasets/pull/8160
- https://github.com/huggingface/dataset-viewer/pull/3358
- https://github.com/huggingface/datasets/issues/8499

## Current limitation

The current TsFile C++ read path is already offset-based. `ReadFile::read()`
accepts an offset and length, and the Reader uses metadata near the end of the
file to locate the required chunks and pages.

However, the Python Reader ultimately passes its input to the C++ `ReadFile`
implementation as a local filesystem path:

```text
TsFileReader("hf://datasets/.../data.tsfile")
    -> Cython/C API passes a path string
    -> C++ ReadFile opens a local path
    -> the local file API cannot open the hf:// path
```

In the Hugging Face TsFile builder, `open()` returns a remote file object
created by `xopen()` in streaming mode. The builder then calls `read(6)` to
read and validate the TsFile magic header. However, this remote file access
capability is not carried into `TsFileReader`.

The missing capability is therefore the connection between a remote
random-access source and the existing offset-based TsFile read path.

## Proposal

Add remote random-access reading support across the TsFile C++ Reader and
Python bindings.

The C++ read path should be able to consume a remote random-access source
provided by a language binding while preserving the existing upper-level file
parsing and query paths.

The Python bindings should connect seekable file sources provided by fsspec or
the calling environment to the C++ read path. Hugging Face and fsspec would
continue to handle URLs, authentication, caching, retries, and HTTP Range
requests.

This proposal covers both the C++ I/O path and the Python integration, without
prescribing a specific internal interface.

The expected behavior is:

1. Existing top-level Reader and query interfaces and semantics remain
   unchanged.
2. The schema, metadata, chunk, page, and query layers do not depend on a
   specific I/O backend.
3. Existing local TsFile reading behavior and performance are unaffected.
4. Selective remote queries transfer only the required file ranges.
5. Both the table model and tree model can use the same remote random-access
   capability.
6. Local and remote reads produce consistent schemas and query results.

## Related local read backend work

The community is already working on an mmap-based local read backend:

- https://github.com/apache/tsfile/issues/903
- https://github.com/apache/tsfile/pull/920

Issue #903 and PR #920 focus on local file access methods and performance,
while this proposal focuses on remote random-access sources provided by the
Python environment.

Both should preserve the existing upper-level read path. Their underlying I/O
work should therefore be coordinated to avoid introducing duplicated or
independent read paths.

## Expected outcome

After this capability is available, the Hugging Face TsFile builder can use
the remote file access context already provided by `xopen()`:

```text
hf:// URL
    -> Hugging Face xopen
    -> fsspec seekable file source
    -> TsFile Python Reader
    -> existing offset-based TsFile read path
    -> HTTP Range requests
    -> Dataset Viewer output
```

This would allow Dataset Viewer to read and preview remote TsFile datasets,
and provide the technical foundation for officially registering
`format:tsfile`, reindexing existing datasets, and enabling the TsFile format
filter.

ColinLeeo:
## Summary

- introduce a C++ `RandomAccessFile` abstraction while keeping local paths on `ReadFile`
- allow Python `TsFileReader` to consume seekable binary file objects without changing the path-only C API
- preserve caller ownership and cursor state, and propagate remote read failures through the existing error model
- package the private Python bridge and document `fsspec` usage

## Testing

- `./mvnw clean verify -P with-python`
- C++ test binary: 922 passed, 3 skipped
- Python tests: 246 passed
- real Hugging Face `HfFileSystem.open()` comparison: 19,701 rows and 29 columns matched a local copy exactly
- file-object cursor, lifetime, short-read, open-error, query-error, and reader-move coverage

Fixes #930

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ python/tsfile/*so*
 python/tsfile/*dll*
 python/tsfile/*dylib*
 python/tsfile/*.h
+!python/tsfile/python_random_access_file.h
 python/tsfile/*.cpp
 python/tsfile/**/*.cpp
 python/tsfile/**/*so*
--- a/cpp/src/file/random_access_file.cc
+++ b/cpp/src/file/random_access_file.cc
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "file/random_access_file.h"
+
+#include <cstring>
+
+#include "common/tsfile_common.h"
+#include "utils/errno_define.h"
+
+namespace storage {
+
+int validate_tsfile(RandomAccessFile& file, unsigned char* file_version) {
+    static const int64_t MIN_FILE_SIZE = 2 * MAGIC_STRING_TSFILE_LEN + 1;
+    if (!file.is_opened()) {
+        return common::E_FILE_READ_ERR;
+    }
+    if (file.file_size() < MIN_FILE_SIZE) {
+        return common::E_TSFILE_CORRUPTED;
+    }
+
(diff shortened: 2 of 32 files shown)


ColinLeeo:
Superseded by #939, which uses the renamed branch without the codex prefix.

PR #949, opened 2026-09-09
Feat/go table model api

ColinLeeo:
## Summary

Add a table-model-only Go API backed by the public C ABI.

The Go writer now binds and retains a single `TableSchema` during construction. Callers can write targetless Tablets or Arrow Record/Table batches, then query table data through one configurable `Reader.Query` entry point.

## Changes

- Add `NewWriter(path, schema, options...)` with schema ownership and validation.
- Change `NewTablet` to accept only columns and row capacity.
- Add Tablet and Arrow batch writes for all supported table data types.
- Add unified query options for time range, Tag filters, pagination, and Arrow batch size.
- Add typed row access with one-based ResultSet column indexes.
- Add Arrow RecordBatch/Table result consumption with `io.EOF` handling.
- Add table schema lookup and enumeration.
- Preserve BLOB length, embedded NUL bytes, empty values, and NULL values.
- Normalize table and column identifiers consistently with the native table model.
- Truncate existing files when creating a writer.
- Remove tree-model symbols and examples from the public Go API.
- Strengthen C ABI argument validation and resource ownership handling.

## Validation

- `go test ./...`
- `go test -race ./tsfile`
- `go vet ./...`
- C++ test suite: 934 passed, 3 skipped because external compatibility fixtures were not configured
- `./mvnw -P with-cpp -DskipTests -Dxml-format.skip=true package`


docs/superpowers/specs/2026-09-08-go-table-api-design.md
HTHou, line 1: superpowers?

cpp/src/cwrapper/tsfile_cwrapper.cc
```diff
@@ -302,14 +366,33 @@ ERRNO tsfile_reader_close(TsFileReader reader) {
...
+            }
+            measurement_list.emplace_back(std::move(name));
+            data_type_list.push_back(
+                static_cast<common::TSDataType>(*(data_types + i)));
+        }
+        return new (std::nothrow)
+            storage::Tablet(measurement_list, data_type_list, max_rows);
+    } catch (...) {
+        return nullptr;
     }
```
jt2594838, line 418: Return ERRNO?

go/tsfile/cgo_bridge.go
```diff
@@ -476,122 +479,89 @@ func (h *tabletHandle) addString(row, column int, value string) error {
...
-	return newError("register device", cerrno(C.tsfile_generic_writer_register_device(
-		C.TsFileGenericWriter(h.ptr), native)))
+	return &writerHandle{
+		nativeHandle: nativeHandle{ptr: unsafe.Pointer(h), release: releaseWriter},
+		file:         unsafe.Pointer(file),
+	}, nil
 }
 
-func (h *writerHandle) registerTable(schema TableSchema) error {
+func allocTableSchema(schema TableSchema) (*C.TableSchema, func(), error) {
```
jt2594838, line 558: Add some comments to explain what the second return value is?

go/tsfile/schema.go
```diff
@@ -79,32 +60,43 @@ func validDataType(value DataType) bool {
...
-}
-
-func validColumnCategory(value ColumnCategory) bool {
-	return value >= ColumnCategoryTag && value <= ColumnCategoryTime
-}
-
-func validateTimeseriesSchema(op string, schema TimeseriesSchema) error {
-	if err := validateCString(op, "timeseries name", schema.Name); err != nil {
-		return err
+func validateAndCopyTableSchema(schema TableSchema) (TableSchema, error) {
```
jt2594838, line 73: Where is the copy needed?

go/tsfile/tablet.go
```diff
@@ -23,35 +23,68 @@ import (
...
 )
 
-// Tablet is a fixed-capacity batch of rows for one device or table.
+// Tablet is a fixed-capacity batch of table rows.
 type Tablet struct {
 	mu      sync.Mutex
 	handle  *tabletHandle
 	columns []TabletColumn
 	maxRows int
+	timeSet []bool
```
jt2594838: Why would we need this?
We do not allow any holes in the time column of a Tablet.

go/tsfile/writer.go
```diff
@@ -79,78 +89,189 @@ func (w *Writer) withHandle(fn func(*writerHandle) error) error {
...
-			return fmt.Errorf("column %d: %w", i, err)
+	tablet.mu.Lock()
+	defer tablet.mu.Unlock()
+	if tablet.handle == nil || tablet.handle.ptr == nil {
+		return ErrClosed
+	}
+	if len(tablet.columns) != len(w.schema.Columns) {
+		return fmt.Errorf("%w: Tablet has %d columns; writer schema has %d",
+			ErrInvalidSchema, len(tablet.columns), len(w.schema.Columns))
+	}
```
jt2594838: Why would a tablet need a mutex?

cpp/src/cwrapper/tsfile_cwrapper.cc
ColinLeeo: Fixed in `eee4c5c93`. `tablet_new` now reports the exact failure code through an `ERRNO*` out parameter, and Go preserves it.

go/tsfile/cgo_bridge.go
ColinLeeo: Added. The second value is the cleanup callback for the C-owned `TableSchema` and its nested allocations.

go/tsfile/schema.go
ColinLeeo: The writer retains the normalized schema for later Tablet validation, so it needs an owned copy. Added a comment to clarify this.

go/tsfile/tablet.go
ColinLeeo: Removed the Go-only timestamp presence tracking and hole validation. Go now follows the native Tablet row semantics.

go/tsfile/writer.go
ColinLeeo: Removed. Tablet lifecycle is caller-managed, so `Close` must not overlap with active writes; no Go-side Tablet mutex is added.

jt2594838 (approved):

PR #963, opened 2026-09-17
Move type dispatch outside batch read and write loops

jt2594838:
Move type selection out of hot per-value loops in TsFile's read, write, and import paths. Pages and non-aligned Tablet columns use generated type-specific batch loops; aligned writes and query projections reuse resolved types. Arrow, Parquet, and TabletBuilder populate column arrays directly and resolve conversion metadata once per column.

The implementation preserves decoder consumption for filtered/deleted rows, pagination accounting, null bitmaps, late-column page alignment, scalar encoding/SDT behavior, and the successfully written timestamp prefix when an out-of-order row is rejected. Regression tests cover these boundaries, supported data types, and DATE representations.

### Validation

- 93 distinct tests passed across `PageDataBatchReaderTest`, `AlignedNullWriteTest`, `TabletColumnWriteTest`, `TabletBuilderTest`, `ValueConverterTest`, `ArrowSourceReaderTest`, and `ParquetSourceReaderTest`; the configured Maven lifecycle executed the selected suites twice, with no failures, errors, or skips.
- Java Spotless checks and `git diff --check` passed.
- Both compared revisions compiled independently; 114 benchmark output hashes matched. The full repository test suite was not run locally.

### Benchmarks

Compared this branch at `5d15440e831dfeb00f025d8c78c716945cc250c2` with `develop` at `8fdbc49bc80eb4a4d6cf6e46cdf0cabd1a1c0837`: JMH 1.37, Temurin 17.0.20.1, Windows 11, Intel i9-12900, one thread pinned to logical CPU 2, 512 MiB heap, 4096 rows, and 3 forks. Initial runs used 5 × 500 ms warmup and measurement iterations; six methods were repeated with 10 × 1 s warmup and 7 × 1 s measurement iterations, retaining both runs.

| Path | Observed time reduction in cases with separated 99.9% intervals |
|---|---:|
| Page reading / column construction | 12.9%–30.2% |
| Null writing / catch-up | 23.8%–33.0% |
| TabletBuilder import | 40.9%–42.1% |
| Arrow reading | 57.8%–60.5% |
| Tablet writing | 16.2%–25.5% |
| Query projection | 3.0%–13.3% |

Of 30 paired scenarios, 25 had lower measured time with separated intervals; five remain inconclusive. Parquet and single-destination scalar projection had overlapping intervals. Both `columnSelected` reruns were affected by unequal background load and are explicitly inconclusive, despite higher measured means. Other IoTDB workloads were active on the machine during extended runs, so these are shared-machine microbenchmark observations, not isolated causal estimates or end-to-end throughput claims. Interval separation is a conservative screening rule, not a paired statistical test.

Benchmark sources, raw results, the detailed report, and reproduction commands remain local under `target/branch-vs-develop/`, as requested, and are not part of this PR. The benchmark report's validation section describes the earlier benchmark phase; the focused unit tests above were run subsequently for PR submission.


jt2594838 (commented):
Implementation rationale and compatibility notes for the submitted changes. Focused tests and benchmark limitations are summarized in the PR description.

java/tools/src/main/java/org/apache/tsfile/tools/ArrowSourceReader.java
```diff
@@ -182,23 +182,15 @@ public SourceBatch readBatch() {
...
-        Object[] row = new Object[numCols];
-        for (int c = 0; c < numCols; c++) {
-          String colName = schemaColumnNames.get(c);
-          FieldVector vec = vectorMap.get(colName);
-          if (vec == null || vec.isNull(r)) {
-            row[c] = null;
-          } else {
-            row[c] = extractValue(vec, r);
-          }
+      Object[][] columns = new Object[numCols][rowCount];
```
jt2594838, line 185: Build SourceBatch columns directly because both Arrow and the destination are columnar. Resolving each vector once removes repeated name lookups and the intermediate row arrays/transpose; absent vectors remain null. Existing Arrow reader tests and the cross-version output checks passed.

java/tools/src/main/java/org/apache/tsfile/tools/ArrowSourceReader.java
```diff
@@ -293,40 +285,75 @@ private List<String> getSchemaColumnNames() {
...
-      return ((Float8Vector) vec).get(row);
-    } else if (vec instanceof BitVector) {
-      return ((BitVector) vec).get(row) != 0;
-    } else if (vec instanceof VarCharVector) {
-      byte[] bytes = ((VarCharVector) vec).get(row);
-      return new String(bytes, StandardCharsets.UTF_8);
-    } else if (vec instanceof VarBinaryVector) {
-      return ((VarBinaryVector) vec).get(row);
+  /** Select the physical vector type once for the entire column. */
+  private void readColumn(FieldVector vec, Object[] output) {
```
jt2594838, line 289: Bind the physical vector type before iterating through its values. Each typed loop retains null checks, native timestamp precision, date conversion, UTF-8 decoding, and the generic fallback, while avoiding an instanceof chain for every cell.

java/tools/src/main/java/org/apache/tsfile/tools/ParquetSourceReader.java
```diff
@@ -164,34 +165,29 @@ public SourceBatch readBatch() {
       Map<String, Integer> parquetColIndex = buildParquetColumnIndex();
 
       int numCols = schemaColumnNames.size();
-      List<Object[]> rows = new ArrayList<>((int) rowCount);
-
-      for (long r = 0; r < rowCount; r++) {
+      int count = Math.toIntExact(rowCount);
+      Object[][] columns = new Object[numCols][count];
```
jt2594838, line 169: Keep the Parquet record reader row-oriented but write directly to the final column arrays. Column indices and extractors are resolved once per batch; missing or malformed fields retain the existing null behavior. This removes the row transpose without changing the input reader protocol.

java/tools/src/main/java/org/apache/tsfile/tools/ParquetSourceReader.java
```diff
@@ -281,37 +277,30 @@ private Map<String, Integer> buildParquetColumnIndex() {
     return index;
   }
 
-  private Object extractValue(Group group, int fieldIndex) {
+  private Function<Group, Object> valueExtractor(int fieldIndex) {
```
jt2594838, line 280: Choose physical/logical-type conversion once per column, including the dedicated INT96 accessor, rather than inspecting schema metadata for every value. The wrapper still checks field presence per row. Local benchmarks showed lower allocation but inconclusive elapsed-time changes for Parquet.

java/tools/src/main/java/org/apache/tsfile/tools/TabletBuilder.java
```diff
@@ -70,31 +73,54 @@ public Tablet build(SourceBatch batch) {
...
-
-        if (tagDefaults.containsKey(colName)) {
-          tablet.addValue(colName, i, tagDefaults.get(colName));
-          continue;
+    // SourceBatch and Tablet are columnar. Keep schema lookup and type dispatch outside the row
+    // loop.
+    for (int col = 0; col < tableSchema.getColumnSchemas().size(); col++) {
+      IMeasurementSchema colSchema = tableSchema.getColumnSchemas().get(col);
+      String colName = colSchema.getMeasurementName();
+      Type type = Type.fromTsDataType(colSchema.getType());
```
jt2594838, line 83: Traverse columns after establishing sorted timestamps so schema lookup, the destination array, and type resolution are reused across rows. Direct insertion preserves Tablet bitmap semantics by unmarking only non-null converted values; tag defaults and sorted source indices retain their original behavior. Conversion is initialized lazily so empty/all-null columns do not invoke converters. The new tests compare this path with the public Tablet insertion API.

java/tools/src/main/java/org/apache/tsfile/tools/ValueConverter.java
```diff
@@ -150,6 +151,22 @@ private static Object fromString(
         .convert(value, isMeasurement, timePrecision);
   }
 
+  /** Resolve both input representations once for a column whose target type is fixed. */
+  static Function<Object, Object> converterFor(
```
jt2594838, line 155: Resolve the string and object conversion services once for a fixed target column. The returned converter still selects the appropriate input representation for each cell, so columns mixing strings and native Java values keep the same conversion rules and timestamp precision.

java/tsfile/src/main/java/org/apache/tsfile/utils/TypeServices.java
```diff
@@ -22,21 +22,151 @@
...
 import org.apache.tsfile.write.chunk.ValueChunkWriter;
 
+import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.function.LongPredicate;
 
 public final class TypeServices {
 
+  /** Type-specialized loops avoid a polymorphic reader invocation for every decoded value. */
+  public static final TypeService<PageDataBatchReader> READ_PAGE_BATCH_SERVICE =
```
jt2594838, line 59: Expose one type-specialized reader per physical representation so callers dispatch once for a batch. Existing scalar services remain available, including as regression-test references; DATE has its own implementation to preserve its special output representations.

java/tsfile/src/main/java/org/apache/tsfile/utils/TypeServices.java
```diff
@@ -22,21 +22,151 @@
...
+            case FLOAT -> FloatPageDataReader.INSTANCE;
+            case DOUBLE -> DoublePageDataReader.INSTANCE;
+            case TEXT, BLOB, STRING, OBJECT -> BinaryPageDataReader.INSTANCE;
+            case ROW, UNKNOWN, VECTOR ->
+                throw new UnSupportedDataTypeException(String.valueOf(type.getTypeEnum()))
+                    .setChecked(true);
+          };
+
+  /** Resolve the Tablet column type before entering its row loop. */
+  public static final TypeService<TabletColumnWriter> WRITE_TABLET_COLUMN_SERVICE =
```
jt2594838, line 75: Select a generated typed writer before the Tablet row loop. The exhaustive type mapping and service checks cover supported aliases and deliberate unsupported-type failures; the shared write context keeps timestamp validation independent of the value representation.

java/tsfile/src/main/codegen/templates/PageDataReaderTemplate.ftl
```diff
@@ -0,0 +1,205 @@
...
+import java.util.function.LongPredicate;
+
+/** Generated from PageDataReaderTemplate.ftl. Dispatch once per batch, not once per value. */
+public final class ${name}PageDataReader implements PageDataBatchReader {
+  public static final ${name}PageDataReader INSTANCE = new ${name}PageDataReader();
+
+  private ${name}PageDataReader() {}
+
+  @Override
+  public void readBatch(Decoder timeDecoder, ByteBuffer timeBuffer, Decoder decoder,
```
jt2594838, line 60: Generate the non-aligned BatchData loop with direct typed decoder/filter/output calls. Values are consumed before deletion and filter checks, preserving decoder position while removing the per-value polymorphic service invocation.

java/tsfile/src/main/codegen/templates/PageDataReaderTemplate.ftl
```diff
@@ -0,0 +1,205 @@
...
+      ${primitive} value = decoder.read${suffix}(buffer);
+      if (!isDeleted.test(timestamp)
+          && (allSatisfy || filter.satisfy${filterSuffix}(timestamp, value))) {
+        output.put${suffix}(timestamp, value);
+      }
+    }
+  }
+
+  @Override
+  public void readAlignedBatch(long[] timestamps, byte[] bitmap, Decoder decoder,
```
jt2594838, line 74: Aligned streams encode only present values. Checking the bitmap before decoding preserves that layout; present but deleted/filtered values are still consumed. Scalar-versus-batch tests exercise nulls, deletions, filters, and both output orders.

java/tsfile/src/main/codegen/templates/PageDataReaderTemplate.ftl
```diff
@@ -0,0 +1,205 @@
...
+      ${primitive} value = decoder.read${suffix}(buffer);
+      if (!isDeleted.test(timestamp)
+          && (allSatisfy || filter.satisfy${filterSuffix}(timestamp, value))) {
+        output.put${suffix}(timestamp, value);
+      }
+    }
+  }
+
+  @Override
+  public long readBlock(Decoder timeDecoder, ByteBuffer timeBuffer, Decoder decoder,
```
jt2594838, line 91: Keep deletion, filter counting, offset, and limit handling in the specialized loop. Both streams are consumed for the stop row just as in the scalar implementation; deleted rows do not increment the filter-rejection count. Regression tests compare output, decoder positions, and pagination state.

java/tsfile/src/main/codegen/templates/PageDataReaderTemplate.ftl
```diff
@@ -0,0 +1,205 @@
...
+      times.writeLong(timestamp);
+      values.write${suffix}(value);
+      builder.declarePosition();
+      pagination.consumeLimit();
+    }
+    return filtered;
+  }
+
+  @Override
+  public void readValues(long[] timestamps, byte[] bitmap, Decoder decoder, ByteBuffer buffer,
```
jt2594838, line 124: Construct typed primitive wrappers within the generated loop and preserve null/deleted positions in the result array. DATE explicitly retains its DATE type marker instead of becoming a plain INT32 wrapper.

java/tsfile/src/main/codegen/templates/PageDataReaderTemplate.ftl
```diff
@@ -0,0 +1,205 @@
...
+      <@rangeColumn dateBinary=true />
+      return;
+    }
+    </#if>
+    <@rangeColumn dateBinary=false />
+  }
+}
+</#list>
+
+<#macro selectedColumn dateBinary>
```
jt2594838, line 169: Selection and deletion must not skip encoded values: every present value is decoded, then selected deleted rows append null while unselected rows append nothing. The DATE/BinaryColumnBuilder variant uses writeDate to retain its representation. Tests include bitmap-byte boundaries and fully discarded selections.

java/tsfile/src/main/codegen/templates/PageDataReaderTemplate.ftl
```diff
@@ -0,0 +1,205 @@
...
+    if (builder instanceof BinaryColumnBuilder binaryBuilder) {
+      <@selectedColumn dateBinary=true />
+      return;
+    }
+    </#if>
+    <@selectedColumn dateBinary=false />
+  }
+
+  @Override
+  public void readColumn(int start, int end, byte[] bitmap, Decoder decoder, ByteBuffer buffer,
```
jt2594838, line 150: Consume present values before the requested range, then append exactly the requested rows including nulls. Keeping the DATE binary-builder branch outside the loop avoids per-row representation dispatch while preserving range semantics.

java/tsfile/src/main/codegen/templates/TabletColumnWriterTemplate.ftl
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.tsfile.write.chunk.TabletWriteContext;
+
+/** Generated typed loop; retain scalar encoding to preserve page boundaries and SDT state. */
+public final class ${name}TabletColumnWriter implements TabletColumnWriter {
+  public static final ${name}TabletColumnWriter INSTANCE = new ${name}TabletColumnWriter();
+
+  private ${name}TabletColumnWriter() {}
+
+  @Override
+  public void write(ChunkWriterImpl writer, long[] times, Object values, BitMap nulls,
```
jt2594838, line 48: Cast the value array once and run a typed row loop, but retain scalar writer.write calls. This preserves existing page boundaries and SDT state that a different bulk-encoding API could change. Nulls are skipped before timestamp validation; successful timestamps advance only after the write. DATE supports both int[] and LocalDate[] inputs.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/page/PageReader.java
```diff
@@ -161,21 +158,16 @@ public BatchData getAllSatisfiedPageData(boolean ascending) throws IOException {
...
-          valueDecoder,
-          valueBuffer,
-          recordFilter,
-          pageData,
-          timestamp,
-          allSatisfy,
-          deletePredicate);
-    }
+    getBatchReader()
+        .readBatch(
```
jt2594838, line 162: Delegate the entire BatchData decode loop to the cached batch reader. The surrounding decompression, filter fast path, BatchData construction, and final flip remain with PageReader, preserving ascending/descending result behavior.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/page/PageReader.java
```diff
@@ -195,36 +187,32 @@ public TsBlock getAllSatisfiedData(LongConsumer filterRowsRecorder) throws IOExc
...
-              paginationController);
-      if (status == PageDataReadStatus.FILTERED) {
-        allFilteredRows++;
-      } else if (status == PageDataReadStatus.STOP) {
-        break;
-      }
-    }
+    long allFilteredRows =
+        getBatchReader()
+            .readBlock(
```
jt2594838, line 193: Delegate TsBlock decoding and pagination as one typed batch operation, then report the returned filter-rejection count through the existing recorder. This removes the per-value status return and dispatch while retaining PageReader ownership of the output builder.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/page/ValuePageReader.java
```diff
@@ -128,19 +126,9 @@ public BatchData nextBatch(long[] timeBatch, boolean ascending, Filter filter)
...
-    for (int i = 0; i < timeBatch.length; i++) {
-      if (((bitmap[i / 8] & 0xFF) & (MASK >>> (i % 8))) == 0) {
-        continue;
-      }
-      long timestamp = timeBatch[i];
-      batchDataValueReader.read(
-          valueDecoder, valueBuffer, filter, pageData, timestamp, allSatisfy, deletePredicate);
-    }
+    getBatchReader()
+        .readAlignedBatch(
```
jt2594838, line 130: Move the aligned BatchData loop behind the cached batch service, passing the bitmap and deletion predicate unchanged. The specialized implementation owns value-stream advancement, including present values discarded by filtering.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/page/ValuePageReader.java
```diff
@@ -167,17 +155,8 @@ public TsPrimitiveType[] nextValueBatch(long[] timeBatch) throws IOException {
...
-              Type.fromTsDataType(dataType));
-    }
-    for (int i = 0; i < size; i++) {
-      if (((bitmap[i / 8] & 0xFF) & (MASK >>> (i % 8))) == 0) {
-        continue;
-      }
-      valueBatch[i] = valueReader.read(valueDecoder, valueBuffer, timeBatch[i], deletePredicate);
-    }
+    getBatchReader()
+        .readValues(timeBatch, bitmap, valueDecoder, valueBuffer, valueBatch, deletePredicate);
```
jt2594838, line 159: Use the same cached batch service for primitive-array output so the type-dependent reader is not invoked for every present row. The surrounding empty-page handling and output array shape remain in ValuePageReader.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/page/ValuePageReader.java
```diff
@@ -193,20 +172,15 @@ public void writeColumnBuilderWithNextBatch(
...
-        if (keepCurrentRow[i]) {
-          columnBuilder.appendNull();
-        }
-        continue;
-      }
-      columnBuilderValueReader.read(
-          valueDecoder, valueBuffer, columnBuilder, keepCurrentRow[i], isDeleted[i]);
-    }
+    getBatchReader()
+        .readColumn(
```
jt2594838, line 176: The selected-column overloads now delegate whole loops with optional deletion masks. Passing a null deletion mask represents the no-deletions overload; null-page handling remains in the caller. The scalar comparison tests cover both forms and keep/deletion combinations.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/page/ValuePageReader.java
```diff
@@ -243,23 +206,15 @@ public void writeColumnBuilderWithNextBatch(
...
-      }
-    }
-    for (int i = readStartIndex; i < readEndIndex; i++) {
-      if (((bitmap[i / 8] & 0xFF) & (MASK >>> (i % 8))) == 0) {
-        columnBuilder.appendNull();
-        continue;
-      }
-      columnBuilderValueReader.read(valueDecoder, valueBuffer, columnBuilder, true, false);
+    getBatchReader()
+        .readColumn(readStartIndex, readEndIndex, bitmap, valueDecoder, valueBuffer, columnBuilder);
```
jt2594838, line 210: Delegate range insertion to the typed reader while keeping the empty-page appendNull fast path. The generated loop consumes the encoded prefix before appending the requested range, preserving subsequent decoder position.

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/AlignedChunkGroupWriterImpl.java
```diff
@@ -209,6 +207,9 @@ public int write(Tablet tablet, int startRowIndex, int endRowIndex)
         emptyValueChunkWriters.add(entry.getValue());
       }
     }
+    // Resolve lazily on the first row to preserve timestamp validation and late-column catch-up.
+    ValueChunkWriter[] writers = new ValueChunkWriter[measurementSchemas.size()];
```
jt2594838, line 211: Cache each column writer and Type for the Tablet call, resolving them lazily on the first accepted row. Lazy resolution preserves timestamp-validation order and the catch-up behavior of newly added columns; the outer row loop continues to synchronize aligned page boundaries.

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/AlignedChunkGroupWriterImpl.java
```diff
@@ -285,17 +290,13 @@ public void tryToAddEmptyPageAndData(ValueChunkWriter valueChunkWriter) throws I
     }
 
     // add empty data of currentPage
-    for (long i = 0; i < timeChunkWriter.getPageWriter().getStatistics().getCount(); i++) {
-      valueChunkWriter.write(0, 0, true);
-    }
+    valueChunkWriter.writeNull(timeChunkWriter.getPageWriter().getStatistics().getCount());
```
jt2594838, line 293: Catch up a newly added column with one null-append call for the current time page. Nulls have no encoded values or value statistics, so this avoids repeated scalar writes while retaining sealed-page catch-up in the existing surrounding logic.

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/AlignedChunkGroupWriterImpl.java
```diff
@@ -285,17 +290,13 @@ public void tryToAddEmptyPageAndData(ValueChunkWriter valueChunkWriter) throws I
...
   }
 
   private void writeEmptyDataInOneRow(List<ValueChunkWriter> valueChunkWriterList) {
     for (ValueChunkWriter valueChunkWriter : valueChunkWriterList) {
-      EmptyValueChunkWriter emptyValueWriter =
-          TypeServices.WRITE_EMPTY_VALUE_TO_CHUNK_SERVICE.call(
-              Type.fromTsDataType(valueChunkWriter.getDataType()));
-      emptyValueWriter.write(valueChunkWriter);
+      // Keep row-wise page-size checks and synchronized page boundaries across all columns.
+      valueChunkWriter.writeNull(1);
```
jt2594838, line 299: Missing measurements append one null directly instead of resolving a type-specific empty-value service per row. The row-wise page-size checks are intentionally retained so all aligned columns continue to seal together.

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/NonAlignedChunkGroupWriterImpl.java
```diff
@@ -116,20 +118,35 @@ public int write(Tablet tablet, int startRowIndex, int endRowIndex)
...
+      int firstRow = startRowIndex;
+      while (firstRow < endRowIndex && nulls != null && nulls.isMarked(firstRow)) {
+        firstRow++;
+      }
+      // An empty column must not inspect its value array or invoke a value writer.
+      if (firstRow >= endRowIndex) {
+        continue;
+      }
+      TabletWriteContext context =
+          new TabletWriteContext(deviceId, measurementId, lastTimeMap.get(measurementId));
```
jt2594838, line 131: Resolve the typed column writer once after finding the first non-null row. Empty/all-null columns avoid inspecting the value array. The finally block commits only the successfully written timestamp prefix if a later row is rejected, retaining out-of-order recovery semantics without per-point map lookups.

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/TabletWriteContext.java
```diff
@@ -0,0 +1,68 @@
...
+
+package org.apache.tsfile.write.chunk;
+
+import org.apache.tsfile.common.constant.TsFileConstant;
+import org.apache.tsfile.exception.write.WriteProcessException;
+import org.apache.tsfile.file.metadata.IDeviceID;
+import org.apache.tsfile.i18n.Messages;
+
+/** Per-column state shared by the type-specialized Tablet write loops. */
+public final class TabletWriteContext {
```
jt2594838, line 28: Keep last-time validation and accepted-point counting in per-column state shared with the generated loop. An explicit hasLastTime flag allows Long.MIN_VALUE as a valid first timestamp; recordTime runs only after a successful write so failures preserve the accepted prefix.

java/tsfile/src/main/java/org/apache/tsfile/write/chunk/ValueChunkWriter.java
```diff
@@ -164,6 +164,11 @@ public void write(long time, Binary value, boolean isNull) {
     pageWriter.write(time, value, isNull);
   }
 
+  /** Appends nulls to the current page without type dispatch, encoding, or statistics updates. */
+  public void writeNull(int count) {
```
jt2594838, line 168: Expose the existing page-level null append through ValueChunkWriter. This advances null positions without type dispatch, encoding dummy values, or updating value statistics, enabling both row-null insertion and late-column catch-up.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/block/SingleDeviceTsBlockReader.java
```diff
@@ -274,6 +275,8 @@ public static class SingleMeasurementColumnContext extends MeasurementColumnCont
 
     private final String columnName;
     private final List<Integer> posInResult;
+    // A measurement context keeps one data type across all batches, so bind its Type once.
+    private final Type type;
```
jt2594838, line 279: A scalar measurement context has a stable data type across its batches, so resolve Type once when constructing the context and reuse it for every destination. This removes repeated enum-to-Type lookup from the projection loop.

java/tsfile/src/main/java/org/apache/tsfile/read/reader/block/SingleDeviceTsBlockReader.java
```diff
@@ -322,10 +326,10 @@ void fillInto(TsBlock block, int blockRowNum) {
       for (int i = 0; i < vector.length; i++) {
         final TsPrimitiveType value = vector[i];
         final List<Integer> columnPositions = posInResult.get(i);
+        final Type type = value == null ? null : Type.fromTsDataType(value.getDataType());
```
jt2594838, line 329: Resolve the vector element type once before copying it to repeated projection destinations. Resolution remains per element because the element itself carries the type; null elements continue to set null on every destination without type lookup.

java/tsfile/src/test/java/org/apache/tsfile/read/reader/PageDataTestSupport.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.tsfile.utils.ReadWriteForEncodingUtils;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+
+/** Deterministic encoded pages shared by regression tests and the standalone JMH benchmark. */
+public class PageDataTestSupport {
```
jt2594838, line 40: Share deterministic PLAIN-encoded fixtures across regression tests and the local benchmark. Separate aligned and non-aligned streams, reproducible null/selection/deletion masks, and multiple primitive representations make comparisons exercise actual decoding rather than mocked service calls.

java/tsfile/src/test/java/org/apache/tsfile/read/reader/PageDataBatchReaderTest.java
```diff
@@ -0,0 +1,272 @@
...
+    TSDataType.FLOAT,
+    TSDataType.DOUBLE,
+    TSDataType.TEXT,
+    TSDataType.STRING,
+    TSDataType.BLOB,
+    TSDataType.OBJECT
+  };
+
+  @Test
+  public void columnsMatchScalarReaders() throws Exception {
```
jt2594838, line 66: Compare the new column loops with the retained scalar service across all supported types, empty/small inputs, bitmap boundaries, selection/deletion combinations, and DATE binary output. This checks representation and null-position compatibility.

java/tsfile/src/test/java/org/apache/tsfile/read/reader/PageDataBatchReaderTest.java
```diff
@@ -0,0 +1,272 @@
...
+            }
+            assertColumn(expected.build(), actual.build());
+          }
+        }
+      }
+    }
+  }
+
+  @Test
+  public void batchesAndPrimitiveArraysMatchScalarReaders() throws Exception {
```
jt2594838, line 113: Compare aligned/non-aligned BatchData and primitive-array outputs with scalar references, including ascending/descending traversal, filters, deletions, and DATE type markers. This guards the separate batch entry points against semantic drift.

java/tsfile/src/test/java/org/apache/tsfile/read/reader/PageDataBatchReaderTest.java
```diff
@@ -0,0 +1,272 @@
...
+        } else {
+          assertEquals(expected.getDataType(), actual[i].getDataType());
+          assertEquals(expected.getValue(), actual[i].getValue());
+        }
+      }
+    }
+  }
+
+  @Test
+  public void blockPaginationPreservesFilterCountsAndDecoderPositions() throws Exception {
```
jt2594838, line 161: Check pagination beyond visible output: limits/offsets also affect filter counts and how far both decoder streams advance. Comparing those states with the scalar implementation catches stop-row consumption regressions.

java/tsfile/src/test/java/org/apache/tsfile/read/reader/PageDataBatchReaderTest.java
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  private static ColumnBuilder builder(TSDataType type, int size, boolean binaryDate) {
+    return binaryDate
+        ? new BinaryColumnBuilder(null, size)
+        : Type.fromTsDataType(type).createColumnBuilder(size);
+  }
+
+  @Test
+  public void discardedRowsConsumeValuesAndEmptyPagesAppendNulls() throws Exception {
```
jt2594838, line 231: Exercise fully discarded values and empty pages explicitly. These boundary cases can produce plausible empty output even if a refactored loop fails to consume encoded input or append the required null positions.

java/tools/src/test/java/org/apache/tsfile/tools/TabletBuilderTest.java
```diff
@@ -35,6 +36,110 @@
 
 public class TabletBuilderTest {
 
+  @Test
+  public void testColumnInsertionMatchesPublicTabletApi() {
```
jt2594838, line 40: Compare direct column population with the public Tablet API across supported types and 0/1/35-row batches. Reversed timestamps, default tags, and null markers verify sorting, representations, and bitmap initialization together.

java/tools/src/test/java/org/apache/tsfile/tools/TabletBuilderTest.java
```diff
@@ -35,6 +36,110 @@
...
+          }
+        }
+      }
+      expected.setRowSize(count);
+      assertEquals(expected, actual);
+    }
+  }
+
+  @Test
+  public void testColumnConversionKeepsSortedRowsAndNulls() {
```
jt2594838, line 112: Mix strings and native values in a column with unsorted timestamps and null-format inputs. This specifically checks that a cached converter still chooses per-cell input handling and keeps tag defaults aligned with sorted rows.

java/tsfile/src/test/java/org/apache/tsfile/write/chunk/AlignedNullWriteTest.java
```diff
@@ -0,0 +1,155 @@
...
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.Collections;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+public class AlignedNullWriteTest {
+  @Test
+  public void tabletWriterCacheKeepsLateColumnPagesAndFailureOrder() throws Exception {
```
jt2594838, line 44: Compare cached Tablet writing with scalar writes while adding a column after earlier rows. Byte/page/statistics comparisons and a rejected timestamp guard both late-column catch-up and the order of validation versus writer creation.

java/tsfile/src/test/java/org/apache/tsfile/write/chunk/AlignedNullWriteTest.java
```diff
@@ -0,0 +1,155 @@
...
+      rejected.addValue("rejected", 0, 42);
+      assertThrows(WriteProcessException.class, () -> batch.write(rejected));
+      assertEquals(2, batch.valueChunkWriterMap.size());
+    } finally {
+      config.setMaxNumberOfPointsInPage(oldLimit);
+    }
+  }
+
+  @Test
+  public void lateColumnsAndMissingRowsKeepPageBoundaries() throws Exception {
```
jt2594838, line 97: Force short pages, add every supported type after a sealed page plus a partial page, then write missing and present rows. Verify synchronized boundaries, null bitmaps, and value statistics so direct null appends cannot silently disturb alignment.

java/tsfile/src/test/java/org/apache/tsfile/write/chunk/TabletColumnWriteTest.java
```diff
@@ -0,0 +1,164 @@
...
+import java.time.LocalDate;
+import java.util.Collections;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+public class TabletColumnWriteTest {
+  @Test
+  public void testEncodedPagesMatchScalarForAllTypes() throws Exception {
```
jt2594838, line 46: Compare sealed bytes, page counts, statistics, and accepted-point counts with scalar writing. The test covers all supported types, null masks, a nonzero source offset, both DATE arrays, and SDT settings for numeric types to guard encoding compatibility.

java/tsfile/src/test/java/org/apache/tsfile/write/chunk/TabletColumnWriteTest.java
```diff
@@ -0,0 +1,164 @@
...
+    scalar.sealCurrentPage();
+    batch.sealCurrentPage();
+    assertEquals(dataType.toString(), scalar.getByteBuffer(), batch.getByteBuffer());
+    assertEquals(scalar.getStatistics(), batch.getStatistics());
+    assertEquals(scalar.getNumOfPages(), batch.getNumOfPages());
+    assertEquals(count, context.getPointCount());
+  }
+
+  @Test
+  public void testOutOfOrderKeepsWrittenPrefixAndSkipsNulls() throws Exception {
```
jt2594838, line 127: Verify that null rows do not trigger timestamp validation and that a later out-of-order row preserves the accepted prefix for a subsequent call. Long.MIN_VALUE, all-null ranges, null value arrays, and empty ranges exercise the context and early-exit boundaries.

hongzhi-gao (approved):

hongzhi-gao (approved):

